PR #12, opened 2015-07-04
implement line join style

stanim:
I have an example which I will add later.
linecapjoin_inkscape

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ pdf.txt
 font/Ubuntu-*
 *.0
 *.swp
+**/*.out
+**/*.test
--- a/fpdf.go
+++ b/fpdf.go
@@ -574,6 +574,8 @@ func (f *Fpdf) AddPageFormat(orientationStr string, size SizeType) {
 	// 	Set line cap style to current value
 	// f.out("2 J")
 	f.outf("%d J", f.capStyle)
+	// 	Set line join style to current value
+	f.outf("%d j", f.joinStyle)
 	// Set line width
 	f.lineWidth = lw
 	f.outf("%.2f w", lw*f.k)
@@ -787,6 +789,27 @@ func (f *Fpdf) SetLineCapStyle(styleStr string) {
 		f.capStyle = capStyle
 		if f.page > 0 {
 			f.outf("%d J", f.capStyle)
+		}
+	}
+}
+
+// SetLineJoinStyle defines the line cap style. styleStr should be "miter",
+// "round" or "bevel". The method can be called before the first page
+// is created. The value is retained from page to page.
+func (f *Fpdf) SetLineJoinStyle(styleStr string) {
+	var joinStyle int
+	switch styleStr {
+	case "round":
+		joinStyle = 1
+	case "bevel":
+		joinStyle = 2
+	default:
+		joinStyle = 0
+	}
+	if joinStyle != f.joinStyle {
+		f.joinStyle = joinStyle
+		if f.page > 0 {
+			f.outf("%d j", f.joinStyle)
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


jung-kurt:
Nice feature -- thanks!

marcusatbang:
Should the magic strings (`"miter"`, `"round"` and `"bevel"`) be listed as package constants?

That may apply elsewhere in the project as well...

jung-kurt:
Good point. At the expense of adding a namespace qualifier (for example, `gofpdf.Miter` or `gofpdf.CnMiter`) the compiler can make sure the argument has been entered correctly.

[This gist](https://gist.github.com/jung-kurt/38ed80b64ba7af6e5956) shows that a little bit of type safety could be added while remaining backwards compatible with ordinary strings.

marcusatbang:
Type safe constants at no runtime cost - looks good.

I'm not fond of the `Cn` sigil though. Is there a Go standard for how constants should be named? Part of me wants to use CAPITAL_LETTERS for constants, but I know that's not Go-like.

jung-kurt:
Camel-case seems to be the convention: http://stackoverflow.com/questions/22688906/go-naming-conventions-for-const

The advantage of the `Cn` prefix is that it helps distinguish the name from entry points.

marcusatbang:
How about `JoinStyle_Miter`?

```go
pdf.SetLineJoinStyle(JoinStyle_Miter)
```

jung-kurt:
golint complains: `don't use underscores in Go names; const draw_fill should be drawFill`

I have my share of problems with golint, but I have reached the conclusion that a clean golint report is worth doing things its way.

marcusatbang:
Mumble grumble mutter fine...  :)

marcusatbang:
I'd still favour something specific like `JoinStyleMiter` over the short `CnMiter`. I feel like we're several decades past the point where saving a couple of letters in an identifier makes sense.

jung-kurt:
Agreed -- your point is well taken.

marcusatbang:
I'll try and make sure the FPDI port follows whatever naming convention you decide on.

jung-kurt:
Excellent. Let's go with your `JoinStyleMiter` suggestion and type them like this:
```
type JoinStyle string

const JoinStyleMiter = JoinStyle("miter")

func SetJoinStyle(style JoinStyle) {}
```
As we've seen, this is backwards compatible with ordinary strings. I'll make the changes to the existing API.

marcusatbang:
:+1:

marcusatbang:
And presumably something similar for other magic strings, such as:
```go
func (f *Fpdf) SetDisplayMode(zoomStr, layoutStr string) {
```
```go
func (f *Fpdf) AddPageFormat(orientationStr string, size SizeType) {
```
```go
func (f *Fpdf) SetLineCapStyle(styleStr string) {
```
...etc. It is extremely helpful that they're backwards compatible with `string`.

jung-kurt:
The backwards compatibility is crucial -- the extra bit of type safety wouldn't be worth a breaking change.

I haven't looked at the entire API in this regard, but at least some arguments can be combined: `"F"`, `"D"`, `"FD"` for drawing. I favor making into constants all combinations rather than using concatenation. That is, 
```
DrawStyleFill = DrawStyle("F")
DrawStyleDraw = DrawStyle("D")
DrawStyleFillDraw = DrawStyle("FD")
```
rather than
 ```
DrawStyleFill + DrawStyleDraw
```

stanim:
When I implemented this, I looked at SetLineCapStyle as a blueprint.
As SetJoinStyle
is introduced only a few days ago and probably nobody is using it yet, you
also could be free to implement it another way.
For SetLineCapStyle et alii this is of course too late.

jung-kurt:
All of your work looks fine. The nice thing about changing the argument type from `string` to, say, `JoinStyle` (where `JoinStyle` is defined as `type JoinStyle string`) is that ordinary strings can still be used. Consequently, no existing code should break.

So, no need to change anything yet. Keep up the great work!

PR #14, opened 2015-07-06
Fill

stanim:
I need to be able to draw some complex figures with gofpdf. To enable to draw 'holes' in filled shapes, the current fillDrawOp is too limited. I wanted to support the full PDF specification on fill styles (both non zero winding and even odd rule). I took great care for backwards compatiblity, as "", "F", "D", "DF" and "FD" continue to work as before. Please look at tutorial 32 and its resulting pdf, as helps clarifying the concept.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -858,14 +858,26 @@ func (f *Fpdf) Line(x1, y1, x2, y2 float64) {
 	f.outf("%.2f %.2f m %.2f %.2f l S", x1*f.k, (f.h-y1)*f.k, x2*f.k, (f.h-y2)*f.k)
 }
 
+// fillDrawOp corrects path painting operators
 func fillDrawOp(styleStr string) (opStr string) {
 	switch strings.ToUpper(styleStr) {
-	case "F":
-		opStr = "f"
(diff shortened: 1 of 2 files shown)


jung-kurt:
Nice work, and awesome tutorial!

One quick question before I merge: if the enhanced fill style requires a reader of version greater than 1.3 (the default), there should be a version adjustment something like this:

```
if f.pdfVersion < "1.4" {
    f.pdfVersion = "1.4"
}
```
This would be done for a feature that requires a reader of version 1.4 or later. Adjust the version number accordingly, or skip the step if version 1.3 suffices.

stanim:
Good question. All path painting operators (see DrawPath docstring) are already supported by pdf version 1.3. You can check it on page 152 of https://wwwimages2.adobe.com/content/dam/Adobe/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference13.pdf
So I suppose there is no reason to bump the pdf version.

jung-kurt:
Thanks for looking into that -- I'll proceed with merge.

stanim:
Thanks!

PR #20, opened 2015-07-09
synchronize file names example pdf files with godoc

stanim:
I had difficulty associating the filenames of the generated pdf's with their corresponding godoc examples. So I renamed the output filenames to correspond with the godoc names:
```
addpage.pdf -> Fpdf_AddPage.pdf
```
 If the example had a broader meaning, I added the previous output name to the method.
For example:
```
fill.pdf -> Fpdf_DrawPath_fill.pdf
```
So now it is very easy to find back the output of a godoc example.


jung-kurt:
Nice work. Many thanks.

PR #21, opened 2015-07-10
Fix SetFontUnitSize name so that we can build

swenson:
Build is currently broken


jung-kurt:
Embarrassing slip.

Thanks for correction.

swenson:
Np :)

PR #23, opened 2015-07-11
fix long lines in documentation examples

stanim:
I had difficulty reading examples on godoc as some lines are too long and require horizontal scrolling in textareas as apart of them is hidden. So I splitted the too long lines into two lines.


jung-kurt:
All good -- thank you!

PR #30, opened 2015-07-12
Implement GetAlpha

### LINKED ISSUE #29: GetAlpha needed
For draw2d I need to be able to look up the transparency and blend mode. As in svg and bitmap drawing RGBA colors are used instead of RGB colors as in gopdf. I need to be able to check the current state, to decide if a new call to SetAlpha is necessary. This is also necessary to preserve the blendmode, if I only want to adjust the alpha. I will prepare a pull request.

stanim:
This has been tested successfully with draw2d.
Fixes #29


stanim:
Thanks!

PR #33, opened 2015-08-04
Templating

marcusatbang:
Implementation of templating in the style of [FPDF_TPL](http://manuals.setasign.com/fpdi-manual/the-fpdf-tpl-class/). There are two parts to this: a `Template` interface and a specific implementation `FpdfTpl`. It diverges from the PHP original in that the method `Fpdf.CreateTemplate` uses a callback in which you write the template and then returns a read-only template, which should help ensure type safety and should avoid confusion.

The interface allows other code to insert templates by other mechanisms; the intention is to eventually use this to implement the document-reading functionality of [FPDI](http://manuals.setasign.com/fpdi-manual/the-fpdi-class/). It may have other uses, such as a mechanism for reading SVGs?

Also, I added a few more util methods that I needed, and that may prove useful to others: `PointType.Transform`, `SizeType.Orientation`, `SizeType.ScaleBy`, `SizeType.ScaleWidth`, `SizeType.ScaleHeight`. The PHP version of these made heavy use of optional arguments (pass either an integer or `null` to get different behaviour), so I had to separate the use cases into different methods for the Go port.

Are the test cases sufficient? `go test -cover` reports 85.4% coverage (up of 0.2pp from 85.2%), and I test some edge cases like images within subtemplates. However there are cases the tests definitely don't cover, such as creating a template divorced from a document. I may add more tests.

Apologies that this took so long to fix up. It was mostly implemented months ago, but life has not been generous lately.

### DIFF
--- a/def.go
+++ b/def.go
@@ -160,6 +160,8 @@ type Fpdf struct {
 	page             int                       // current page number
 	n                int                       // current object number
 	offsets          []int                     // array of object offsets
+	templates        map[int64]Template        // templates used in this document
+	templateObjects  map[int64]int             //template object IDs within this document
 	buffer           fmtBuffer                 // buffer holding in-memory PDF
 	pages            []*bytes.Buffer           // slice[page] of page content; 1-based
 	state            int                       // current document state
--- a/doc.go
+++ b/doc.go
@@ -48,6 +48,8 @@ Features
 
 • Layers
 
+• Templates
+
 gofpdf has no dependencies other than the Go standard library. All tests pass
 on Linux, Mac and Windows platforms. Like FPDF version 1.7, from which gofpdf
 is derived, this package does not yet support UTF-8 fonts. However, support is
@@ -193,7 +195,8 @@ implementations for dashed line drawing and generalized font loading. Stani
 Michiels provided support for multi-segment path drawing with smooth line
 joins, line join styles, enhanced fill modes, and has helped greatly with
 package presentation and tests. Bruno Michel has provided valuable assistance
-with the code.
+with the code. Templating is adapted from the FPDF_Tpl library created by
+Jan Slabon and Setasign.
 
 Roadmap
 
--- a/fpdf.go
+++ b/fpdf.go
@@ -72,6 +72,8 @@ func fpdfNew(orientationStr, unitStr, sizeStr, fontDirStr string, size SizeType)
 	f.fonts = make(map[string]fontDefType)
 	f.fontFiles = make(map[string]fontFileType)
 	f.diffs = make([]string, 0, 8)
+	f.templates = make(map[int64]Template)
+	f.templateObjects = make(map[int64]int)
 	f.images = make(map[string]*ImageInfoType)
 	f.pageLinks = make([][]linkType, 0, 8)
 	f.pageLinks = append(f.pageLinks, make([]linkType, 0, 0)) // pageLinks[0] is unused (1-based)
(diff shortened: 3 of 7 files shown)


marcusatbang:
Btw, I did some git trickery to present you with a simplified change request, rather than the full messy history of everything I've been doing. My `master` branch has half-written stuff to do with FPDI in it.

All told this is quite a large change, so I recommend you test it before accepting the pull request.

jung-kurt:
This is a very welcome addition to the project -- nice work! Based on a quick skim of the changes, all the programming, including example, looks excellent. 

The only wrinkle that I see right now is one of licensing. The Apache license is slightly less permissive than the MIT license under which the rest of the package is released. (At least, that is my takeaway from https://en.wikipedia.org/wiki/License_compatibility.) Are you able to change the license? If not, perhaps we can implement a plugin interface with which your templating package could be registered at runtime. In that case, the application would import both gofpdf and your template package, and your package's functionality would become available to gofpdf in the same way that database drivers register themselves with the database/sql package.

marcusatbang:
I had a feeling licensing would be an issue.

I would vastly prefer to switch the license to match the rest of your project allowing it to be integrated without artificial boundaries. But I'm not sure of the conditions under which that works. The original libraries, FPDF_TPL and FPDI, are under Apache 2.0 license; my reading of that (as a programmer, not a lawyer) suggested that derivative works such as a port were permitted provided they preserved the same license as the original.

I contacted [Setasign](http://www.setasign.com/), the original authors of FPDI, before I started and got their permission. Legally speaking that means nothing, but I felt it would be impolite not to; they said to go right ahead. Would the best solution be to go back to them and ask if they'd be willing to release FPDI under an MIT license?

jung-kurt:
It is unlikely the Setasign authors would change their license, but they might give their blessing to letting you change the license of your derived work, especially since the MIT and Apache licenses are so close in nature. Or, they may simply have suggestions about how we resolve this issue.

marcusatbang:
I've asked Setasign what they think.

jung-kurt:
Maybe they would be OK with you releasing your derived work under both the MIT and Apache licenses, the way, for example, jQuery used to be released under dual licenses. That way, we'd have no trouble including your work under the MIT portion.

JanSlabon:
Hi! From our point of view MIT code can be included in ASL 2.0 projects (as described (http://www.apache.org/legal/resolved.html#category-a)[here]. I'm not sure about the other way? Did you found any references like this link from the view of the MIT?

marcusatbang:
Thanks, Jan.

If the ported code was isolated then this would be an easier question, but for the port to work well in Go it needs to be merged in with the rest of the project. I'm not a lawyer, but my understanding is that your blessing is all that's needed to convert the code from MIT to ASL 2.0.

I'll update the files to the Apache license tonight and push, making sure to credit you.

JanSlabon:
I'm not sure if this would be the solution from the view of @jung-kurt ?

marcusatbang:
I've pushed a change with what I believe to be the correct license and attribution (I'm modified `doc.go` to attribute Setasign). Is that enough to satisfy everyone and get this merged?

JanSlabon:
I'm fine with this :+1:

jung-kurt:
Thanks, Marcus. Thanks, Jan. I'll get moving on this.

marcusatbang:
Excellent. Thank you, Jan.

I'll continue porting the rest of FPDI when time allows. So you can expect to see it in the next century or two.

jung-kurt:
Excellent! It will be worth the wait.

jung-kurt:
This is a great module for the gofpdf package. Many thanks, Marcus and Jan, for your excellent work.

PR #34, opened 2015-08-12
Allow the tests to work on Windows

marcusatbang:
The tests currently don't pass on Windows because of a silly little thing:

```
--- FAIL: ExampleFpdf_MoveTo (0.00s)
got:
Successfully generated pdf\Fpdf_MoveTo_path.pdf
want:
Successfully generated pdf/Fpdf_MoveTo_path.pdf
```

etc for all the tests. This tiny change allows the tests to pass.

### DIFF
--- a/fpdf_test.go
+++ b/fpdf_test.go
@@ -109,6 +109,7 @@ func exampleFilename(baseStr string) string {
 
 func summary(err error, fileStr string) {
 	if err == nil {
+		fileStr = filepath.ToSlash(fileStr)
 		fmt.Printf("Successfully generated %s\n", fileStr)
 	} else {
 		fmt.Println(err)


jung-kurt:
Good catch -- it would have been a century or two before I caught *that* one.

stanim:
@jung-kurt 
May I ask which OS do you use?

jung-kurt:
@stanim 
Linux, usually an Ubuntu server via a Chromebook Pixel. I have ready access to a Mac, but Windows is increasingly remote for me. To access it, I have to boot up an aging laptop to Windows XP.

How about you?

stanim:
@jung-kurt
Arch Linux.

marcusatbang:
Linux at work (Ubuntu on servers, Mint on my own laptop, and a few messy old Gentoos we haven't retired yet). Windows at home, but with a lot of cygwin stuff installed so I can use something approaching a real command line.

jung-kurt:
Hmmm...a common theme here.

marcusatbang:
Go fits well with the Unix philosophy, of small tools that do one thing well, so it's no great surprise that a disproportionate number of its programmers would be use Linux.

PR #35, opened 2015-08-13
Register remote image

jelmersnoeck:
This lets you register an image that is hosted somewhere else. This way
you can include an image from another service and you don't have to
store it locally.

I'm fairly new to Go and was just wondering what the best way to run the test was? At the moment I made a package in my go folder `github.com/jung-kurt` and added a symlink to my local repository to run the tests from there.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2370,7 +2370,7 @@ func (f *Fpdf) RegisterRemoteImage(urlStr, tp string) (info *ImageInfoType) {
 	file := bytes.NewReader(body)
 
 	if tp == "" {
-		tp = f.getFileExtensionFromString(urlStr)
+		tp = f.ImageTypeFromMime(resp.Header["Content-Type"][0])
 	}
 
 	return f.RegisterImageReader(urlStr, tp, file)
--- a/fpdf.go
+++ b/fpdf.go
@@ -2340,12 +2340,7 @@ func (f *Fpdf) RegisterImage(fileStr, tp string) (info *ImageInfoType) {
 
 	// First use of this image, get info
 	if tp == "" {
-		pos := strings.LastIndex(fileStr, ".")
-		if pos < 0 {
-			f.err = fmt.Errorf("image file has no extension and no type was specified: %s", fileStr)
-			return
-		}
-		tp = fileStr[pos+1:]
+		tp = f.getFileExtensionFromString(fileStr)
 	}
 
 	return f.RegisterImageReader(fileStr, tp, file)
@@ -2367,12 +2362,7 @@ func (f *Fpdf) RegisterRemoteImage(urlStr, tp string) (info *ImageInfoType) {
 	file := bytes.NewReader(body)
 
 	if tp == "" {
-		pos := strings.LastIndex(urlStr, ".")
-		if pos < 0 {
-			f.err = fmt.Errorf("image file has no extension and no type was specified: %s", urlStr)
-			return
-		}
-		tp = urlStr[pos+1:]
+		tp = f.getFileExtensionFromString(urlStr)
 	}
 
 	return f.RegisterImageReader(urlStr, tp, file)
@@ -2502,6 +2492,15 @@ func (f *Fpdf) Output(w io.Writer) error {
 		f.err = err
 	}
 	return f.err
+}
+
+func (f *Fpdf) getFileExtensionFromString(value string) (extension string) {
+	pos := strings.LastIndex(value, ".")
+	if pos < 0 {
+		f.err = fmt.Errorf("image file has no extension and no type was specified: %s", value)
+		return
+	}
+	return value[pos+1:]
 }
 
 func (f *Fpdf) getpagesizestr(sizeStr string) (size SizeType) {
--- a/fpdf.go
+++ b/fpdf.go
@@ -2355,10 +2355,18 @@ func (f *Fpdf) RegisterRemoteImage(urlStr, tp string) (info *ImageInfoType) {
 		return
 	}
 
-	resp, _ := http.Get(urlStr)
+	resp, err := http.Get(urlStr)
 	defer resp.Body.Close()
 
-	body, _ := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		f.SetError(err)
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		f.SetError(err)
+	}
+
 	file := bytes.NewReader(body)
 
 	if tp == "" {
--- a/fpdf.go
+++ b/fpdf.go
@@ -2362,18 +2362,11 @@ func (f *Fpdf) RegisterRemoteImage(urlStr, tp string) (info *ImageInfoType) {
 		f.SetError(err)
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		f.SetError(err)
-	}
-
-	file := bytes.NewReader(body)
-
 	if tp == "" {
 		tp = f.ImageTypeFromMime(resp.Header["Content-Type"][0])
 	}
 
-	return f.RegisterImageReader(urlStr, tp, file)
+	return f.RegisterImageReader(urlStr, tp, resp.Body)
 }
 
 // GetXY returns the abscissa and ordinate of the current position.
--- a/fpdf.go
+++ b/fpdf.go
@@ -2340,7 +2340,7 @@ func (f *Fpdf) RegisterImage(fileStr, tp string) (info *ImageInfoType) {
 
 	// First use of this image, get info
 	if tp == "" {
-		tp = f.getFileExtensionFromString(fileStr)
+		tp = f.imageTypeFromString(fileStr)
 	}
 
 	return f.RegisterImageReader(fileStr, tp, file)
@@ -2495,7 +2495,7 @@ func (f *Fpdf) Output(w io.Writer) error {
 	return f.err
 }
 
-func (f *Fpdf) getFileExtensionFromString(value string) (extension string) {
+func (f *Fpdf) imageTypeFromString(value string) (extension string) {
 	pos := strings.LastIndex(value, ".")
 	if pos < 0 {
 		f.err = fmt.Errorf("image file has no extension and no type was specified: %s", value)
--- a/fpdf.go
+++ b/fpdf.go
@@ -2350,6 +2350,33 @@ func (f *Fpdf) RegisterImage(fileStr, tp string) (info *ImageInfoType) {
 	return f.RegisterImageReader(fileStr, tp, file)
 }
 
+// RegisterRemoteImage registers a remote image. Downloading the image from the
+// provided URL and adding it to the PDF but not adding it to the page. Use
+// Image() with the same URL to add the image to the page.
+func (f *Fpdf) RegisterRemoteImage(urlStr, tp string) (info *ImageInfoType) {
+	info, ok := f.images[urlStr]
+	if ok {
+		return
+	}
+
+	resp, _ := http.Get(urlStr)
+	defer resp.Body.Close()
+
+	body, _ := ioutil.ReadAll(resp.Body)
+	file := bytes.NewReader(body)
+
+	if tp == "" {
+		pos := strings.LastIndex(urlStr, ".")
+		if pos < 0 {
+			f.err = fmt.Errorf("image file has no extension and no type was specified: %s", urlStr)
+			return
+		}
+		tp = urlStr[pos+1:]
+	}
+
+	return f.RegisterImageReader(urlStr, tp, file)
+}
+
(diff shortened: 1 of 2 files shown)


jung-kurt:
This is a worthwhile enhancement -- thanks. I had thought previously that registering a remote image on the net wasn't worth linking the http package and that it made more sense for the application to do this. But there is enough opportunity for things to go wrong that I believe it makes sense to include it.

Two things. I would like to see the error returned from `http.Get` examined. It is quite reasonable to expect a failure at this point. Also, a url like `http://example.com/foo.jpg?bar=baz` will cause problems with the current method of obtaining the file extension. It would be better to break up the url with `url.Parse` and then process the `Path` member.

stanim:
@jelmersnoeck
Thanks for the proposal. Please don't ignore the errors with underscores as in:

    resp, _ := http.Get(urlStr)
    body, _ := ioutil.ReadAll(resp.Body)

Handle them appropriately: setting them to f.err and return immediately.

I can recommend this blog post for a nice workflow for contributing to open source git repositories in go: https://splice.com/blog/contributing-open-source-git-repositories-go/ If you follow this you can just run:

    go test -cover

jung-kurt:
Valuable link on contributing to Go projects -- thanks!

jelmersnoeck:
Thank you both for the feedback. I'll have a look at refactoring.

@jung-kurt I was also thinking on making a PR for parsing an image from a string (base64). We are currently looking at separating our PDF Generation into it's own service and we'd need to send over some local files/temporary images (like Barcodes). I'm wondering if this is something you'd like to see in this package?

Besides that I'm also looking into optimising this (coming out of necessity of our implementation, we need around 72 images in one PDF, which we need to create on demand). Currently, all the images would get fetched synchronously. If this was implemented with a goroutine all the images could be downloaded simultaneously and then put on the PDF at the end (as happens now). A simple example of this is [here](https://github.com/jelmersnoeck/photo-fetcher/blob/master/photo_fetcher.go).

stanim:
@jung-kurt 
Good point about the extra dependency. Now I think about it, you are right. I am not in favour of merging this pull request, but you decide. As otherwise someone else also wants to add ftp and gofpdf ends up being bloated with many dependencies. I see that there is already a better example for image urls: http://godoc.org/github.com/jung-kurt/gofpdf#ex-Fpdf-RegisterImageReader

@jelmersnoeck
It is great you want to contribute something to this project. However having a closer look to your code, I notice you make a beginner mistake with ioutil.ReadAll. There is no need to read first the resp.Body and turn it back into a reader, as resp.Body is a reader already. See also: https://www.datadoghq.com/blog/crossing-streams-love-letter-gos-io-reader/

jung-kurt:
@stanim 
Good catch with the `ioutil.ReadAll` mistake -- I missed that.

@jelmersnoeck 
The concurrent image loader example is creative, but I wonder if it, like the HTTP image retrieval, is best done by the application. Am I right in assuming the advantage of using goroutines here would be to simply get the resources from one or more remote servers to the local host as quickly as possible? Embedding them into the PDF will still need to wait for all images to be retrieved. Your thoughts, @stanim?

jelmersnoeck:
@stanim thanks for all the information! I've updated it accordingly. I've used the `RegisterImageReader` example as a base instead of `url.Parse` to get the file extension as @jung-kurt mentioned.

I do understand the concerns about the http library and with the provided example (that I seem to have missed), this seems very easy to do. So this could indeed live on the application level, integrated with optimisations to speed up the download.

Regarding concurrent image loading, you would indeed still have to wait for all the images to be retrieved, but the time it takes to do this would equal(roughly) your slowest download instead of the overall time of your downloads.

stanim:
The current code waits for all images to be downloaded before they can embedded in the pdf. This gives you the speed advantage of concurrent downloading, but postpones the pdf generation unnecessarily. A way to further optimize this is to do the pdf generation in a single goroutine to which the images are sent through a channel. (Writing from different goroutines to the same pdf is not safe.) 
* If the order is not important, the images could be pushed in that channel after being downloaded and be embedded immediately in the pdf. 
* If the order is important, the channel element should consist of the image with the index. The pdf generating goroutine can than wait (with a select statement) for previous images if they take longer. This avoids to wait for all images to be downloaded before they can be embedded.

Of course this should be an implementation for the application, not for this library.

That said ... an idea for welcoming contributions, which don't belong in the core library, is to have a `github.com/jung-kurt/gofpdf/contrib` package with various contributed subpackages. This allows us  to have useful functions which take fpdf as an argument, instead of adding them as extra methods and dependencies to gopdf. People can than cherry pick the `contrib` subpackages they want. If after a while a certain contributed function proves to be very useful, it can be migrate as a method to gofpdf core.

This is analogous to django contrib (see http://www.djangobook.com/en/2.0/chapter16.html ).

So instead of adding this method to `github.com/jung-kurt/gofpdf`: 

    func (f *Fpdf) RegisterRemoteImage(urlStr, tp string) (info *ImageInfoType)

This could be implemented as a function in `github.com/jung-kurt/gofpdf/contrib/httpfpdf`: 

    func RegisterRemoteImage(f *gofpdf.Fpdf, urlStr, tp string) (info *gofpdf.ImageInfoType)

stanim:
@jelmersnoeck
Please change this code:

    resp, err := http.Get(urlStr)
    defer resp.Body.Close()

    if err != nil {
        f.SetError(err)
    }

into:

    resp, err := http.Get(urlStr)

    if err != nil {
        f.SetError(err)
    }

    defer resp.Body.Close()

The reason is if `http.Get` fails, `resp` might be nil. Calling `Close` on `resp.Body` will cause a panic.
See: http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

jung-kurt:
@stanim 
> an idea for welcoming contributions, which don't belong in the core library, is to have a github.com/jung-kurt/gofpdf/contrib package with various contributed subpackages

I really like this idea. We may have to expose a little more of the gofpdf internals in order for independent, importable packages to accomplish some tasks, but I think this can be done in a reasonable way.

We should require that all contributions like this contain their own example tests. To make things uniform, it would be desirable to factor the test helper routines and constants into their own package so they could be shared among the gofpdf examples as well as the contribution examples.

jelmersnoeck:
That looks like a neat solution indeed. I'll have a look at refactoring it to follow that pattern.

jelmersnoeck:
@stanim Thanks again for that article, it's very interesting! Am I correct to assume that my function should've included a `return` as well? (At the moment, there is no `return` in there when there's an error). I was also wondering, if an error takes place, should it still be added to the Images map?

As suggested, I've also made the adjustment of moving this to a `contrib` package (I've named it `httpfpdf` as @stanim suggested so other people could add more http related functionality in there). @jung-kurt I'm not sure if you want me to pull those changes into this PR or if you want me to open a new one? The diff can be seen here so far: https://github.com/jung-kurt/gofpdf/compare/master...jelmersnoeck:contrib-package?expand=1

(I'm not entirely sure what to do with the common functions in gofpdf_test and httpfpdf_test though)

jung-kurt:
@jelmersnoeck 
> Am I correct to assume that my function should've included a return as well?

Does this regard error handling? If so, yes, you should return after setting an error. The idea is that once an unrecoverable error occurs, no further document generation should take place.

The `contrib` directory looks good. Let me mull over the idea of exporting the `images` array, however. Instead of exporting it, we may want to want to hide it behind some methods. As more packages are developed in the `contrib` directory, more currently internal fields will need to be accessed, so I would like to arrive at a solution that doesn't simply export everything.

Also, I will think about the best way to handle the common test functions. No need to change your pull request -- I will make any changes in this regard before pushing back to the repository.

Many thanks for your changes.

stanim:
@jung-kurt 
I think the `images` array code:

  +	info, ok := f.Images[urlStr]
+	if ok {
+		return
+	}

stanim:
I think the `images` code:

    info, ok := f.Images[urlStr]
    if ok {
        return
    }

can be dropped. This is for the corner case when some erroneous code tries to register the same image twice. Moreover RegisterImageReader does this check. We should be only expose internals for main behaviour, not for corner cases.

@jelmersnoeck 
In case of an error a return is indeed needed. In idiomatic go functions that can invoke an error, always have an error return parameter. Unfortunately gofpdf follows a different pattern with `Fpdf.SetError`. This is one of few things I regret in gofpdf.

I think when everything looks ok and is ready to be merged, I would prefer if you do a new pull request with a clean commit history. So that you only cause a single commit to gofpdf, as otherwise the commit history gets bloated with unnecessary steps and code. However @jung-kurt might disagree and I will respect his decision.

jung-kurt:
> Unfortunately gofpdf follows a different pattern with Fpdf.SetError. This is one of few things I regret in gofpdf.

I do not regret this. I believe it keeps simple cases simple, and more complex cases possible. I think it captures the spirit of Rob Pike's article about [error handling](https://blog.golang.org/errors-are-values). For one thing, a programmer is still able to use the following idiom

    if err = pdf.Something(); err != nil {
        // process error
    }

by using the construct

    if pdf.Something(); pdf.Err() {
        // process pdf.Error()
    }

Semantically, this isn't exactly the same because `pdf.Error()` may indicate an error that occurred before `pdf.Something()` was called. However, the result would be the same if every opportunity for an error was checked.

For more complex cases, like

    for _, fileStr := range []string{"a.png", "b.png", "c.png"} {
        if err = pdf.Image(fileStr, 10, 6, 30, 0, false, "", 0, ""); err == nil {
            break;
        }
    }

the following, admittedly more cumbersome, construct can be used:

    if pdf.Ok() {
        for _, fileStr := range []string{"a.png", "b.png", "c.png"} {
            if pdf.Image(fileStr, 10, 6, 30, 0, false, "", 0, ""); pdf.Ok() {
                break;
            } else {
                pdf.ClearError(); // Mistakenly omitted until recently
            }
        }
    }

> I would prefer if you do a new pull request with a clean commit history

I agree.

marcusatbang:
Building a PDF page can consist of a _lot_ of small actions. Having an error accumulator, the way gofpdf does it, allows you to check for an error at the end of a bunch of these actions rather than after each individual one.

jelmersnoeck:
@jung-kurt cool, I'll make a new PR with a single commit. Would you like to see it without the `f.Images` block (as @stanim suggested, it could be dropped) or do you want me to wait until you've thought about how to handle the internals?

Out of personal usage experience I'm not entirely convinced to drop it. We currently generate PDFs with 72 remote images on demand. These could all be different, but could also be the same, there is no way to know this(in our case). Having that check in place drops the need to re-download the image and assign it. This might be optimising out of the scope for this PDF generator though.

stanim:
@jelmersnoeck 
Why not wrap the RegisterRemoteImage with a function which checks if it has been downloaded already. This would give the same result and prevents us from exposing unnecessary internals. To keep track of urls it is common in Go to use a map of empty structs, as this consumes the least memory:

    images:= make(map[string]struct{})

    func RegisterRemoteImage(f *gofpdf.Fpdf, urlStr, tp string) {
        _, ok:= images[urlString]
        if ok {
            return
        }
        fhttp.RegisterRemoteImage(f, urlStr, tp)
    }

(I propose here also to use `fhttp` instead of `httpfpdf` to keep it shorter. So the pattern for contributed packages could be `fxxx`.)

If you need the image infos, you can use a map of ImageInfoType instead of empty structs. I think this should be solved at application level, not inside the PDF generator. Anyway I think `f.images` should never be exposed, as a last resort you could add a new method:

    var NoImageInfo ImageInfoType
    func (f *Fpdf) GetImageInfo(urlStr string) (info ImageInfoType) {
        info, ok:= f.images[urlStr]
        if !ok {
            return NoImageInfo
        }
        return info
    }

jung-kurt:
Each of these recommendations sound good to me.

jelmersnoeck:
Perfect. I'll change it to use it's own image map so it doesn't have to touch the internals and open a new PR.

I'll close this PR and link it to the other one once it's in for future references.

stanim:
After a night sleep, maybe `GetImageInfo` is maybe the best option as it has two advantages:
- no external bookkeeping of the image map for other applications
- universal access to image width and height independent whether it comes from a file/url/...

jung-kurt:
@stanim Good points.

I am inclined to use the following signature rather than return a special variable.

    func (f *Fpdf) GetImageInfo(urlStr string) (info ImageInfoType, ok bool) {
            var infoPtr *ImageInfoType
            infoPtr, ok = f.images[urlStr]
            if ok {
                    info = *infoPtr
            }
            return
    }

If this looks good, I'll check it in.

stanim:
Fine for me!
Op 19 aug. 2015 12:52 schreef "Kurt Jung" <notifications@github.com>:

jung-kurt:
[Done](https://github.com/jung-kurt/gofpdf/commit/27b605858db6d8f5439e9b689f70626801324f02)

PR #37, opened 2015-08-19
Register remote image as a package.

jelmersnoeck:
This is a wrapper around manually having to download the image and
registering it. It registers an image based on it's URL, it will
download that image and add it to the PDF. It will however not add it
immediately to the page. This is done by calling `Image()` with the same
URL.

As discussed in https://github.com/jung-kurt/gofpdf/pull/35, this contains the changes within a single commit.


jung-kurt:
@stanim 

> (I propose here also to use fhttp instead of httpfpdf to keep it shorter. So the pattern for contributed packages could be fxxx.)

I agree that we should not include `fpdf` in the contributed project names. However, do we need the `f`? I think project names would be cleaner without it.

@jelmersnoeck 

> (I'm not entirely sure what to do with the common functions in gofpdf_test and httpfpdf_test though)

Let's keep them there for now. When we are satisfied that enough people are using go 1.4 and over, we can use the [internal package](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit) scheme to factor them into their own internal package.

stanim:
@jelmersnoeck 
Thanks for your nice work and patience.

@jung-kurt 
What name would you suggest for this package? I wouldn't use "http" or "image" as that conflicts with the standard library. Using a fprefix has just the advantage to have a clear naming system: f+(name of the main package).
I propose we call the function RegisterHttpImage. (Later maybe somebody adds RegisterFtpImage. RegisterRemoteImage is too general.)

marcusatbang:
> Let's keep them there for now. When we are satisfied that enough people are using go 1.4 and over, we can use the internal package scheme to factor them into their own internal package.

You can create a package called `internal` in older version, it just won't enforce the rules. I vote to move the common test functions now, since contributors are starting to use `contrib`.

stanim:
Using `internal` for common test functions seems a good strategy to me.

@marcusatbang 
Please use the new GetImageInfo method to check if the image is already registered to prevent downloading the same file multiple times.

@jung-kurt 
I would propose to change this:

    func (f *Fpdf) GetImageInfo(imageStr string) (info ImageInfoType, ok bool)

into this:

    func (f *Fpdf) GetImageInfo(imageStr string) (info *ImageInfoType)

which returns `nil` if the image is not registered yet. I find this more idiomatic.

jung-kurt:
@stanim 

Returning a pointer, possibly nil, is fine in this case. In cases in which structure fields are exported, for example to work with reflection or encoding packages, I prefer returning the structure itself so external packages can not modify internals. I'll make your proposed change.

stanim:
@jung-kurt 
I agree, but RegisterImage already returned that pointer:

    func (f *Fpdf) RegisterImage(fileStr, tp string) (info *ImageInfoType)

so it makes it more easy for contributed RegisterImage functions to do the same.

jung-kurt:
@stanim Good point. [Done](https://github.com/jung-kurt/gofpdf/commit/1800f941981bc8ffc3229ba79304ca76974767de)

jelmersnoeck:
@stanim I've updated it to use the `GetImageInfo` function. I've also renamed it to `RegisterHttpImage`.

@jung-kurt do you want to keep it `fhttp` or would you prefer a different name?

jung-kurt:
@jelmersnoeck Nice work. Regarding the name, I think something like `httpimg` or `webimg` would be descriptive since the helper routine involves both. Then I think you could reduce stutter by calling the principal method `webimg.Register()` rather than `webimg.RegisterHttpImage()`. However, I will await comments by @stanim since he catches a lot of stuff that I miss.

jelmersnoeck:
I was actually just wondering the same. With the example @stanim gave, it would maybe belong in a `ftpimg` package (and this one in - as you mentioned - a `httpimg`). Both would then just have a `Register()` function.

stanim:
`httpimg.Register()` would be my preference. Later there can be `ftpimg` or maybe even `httpfont`.

jung-kurt:
@stanim Good -- let's proceed with that.

jelmersnoeck:
Cool, I'll refactor this tonight and squash it all into 1 commit.

jelmersnoeck:
@jung-kurt @stanim I've renamed the package to `httpimg` and updated the function to `Register()`. If you're happy with this I'll squash it into one commit.

stanim:
@jelmersnoeck 
Thanks for your patience ;-) I think you should rename `ExampleFpdf_AddHttpImage()` to `ExampleRegister()` in order to make it work for godoc. Please run and check the documenation with:

    godoc -http=:6060

I also would propose to rename `"Fpdf_AddHttpImage"` to `contrib_httpimg_Register"`. Or @jung-kurt, would you prefer it contrib pdfs go in a `contrib/pdf` subfolder? I would prefer also if there is a cleanup function for contrib generated pdfs, see: https://github.com/jung-kurt/gofpdf/blob/master/fpdf_test.go#L44

jelmersnoeck:
@stanim good point about the cleanup! I wasn't sure to copy over more as @jung-kurt mentioned looking into making this more available for contribution packages. I've added that now.

I've renamed the file as well. Just to be clear, at the moment this writes it to it's own folder (so `contrib/httpimg/pdf`).

stanim:
I think I prefer to have the generated pdfs central in `jung-kurt/gofpdf` or `jung-kurt/gofpdf/contrib`. As than someone can just run:

    go test ./...

And quickly compare/inspect the generated pdfs in one folder, rather than scout them in different folders. @jung-kurt What do you think? 

We really need to get this right, as this will be the example structure for future contribs.

jung-kurt:
@jelmersnoeck All these hoops we're having you jump through are because your package is the first to be placed in the `contrib` directory and we want to get it right. Thanks for bearing with us! Hold off just a little longer and I will commit an example helper package using the `internal` convention.

@stanim Aesthetically, I like the idea of each contributed package keeping all changes (generated PDFs, for example) in its own directory. However, this being the go language, each contributed package's import path clearly defines where in the gofpdf tree it belongs, so it is not like it can be imported to a location in which the ancestor path is different than what we expect it to be. One problem, not a big one, with putting the example PDFs in the main pdf directory is that they will be removed when fpdf_test is run. This could potentially be unexpected. If we opt for a `contrib/pdf`, we face a similar problem: who is responsible for cleanup. Would each package clean up only its own PDFs? I guess I vote for putting them into the main `pdf` directory and accepting a slight impedance mismatch between generation and cleanup.

jung-kurt:
@stanim 

> And quickly compare/inspect the generated pdfs in one folder, rather than scout them in different folders.

Is there a chance that gofpdf_test will run after one or more contrib tests and prematurely remove the PDFs?

jung-kurt:
I just committed a change () in which `exampleFilename()` (now `example.Filename()`), `summary()` (now `example.Summary()`) and the various font, image and test helper routines have been placed in their own package (`internal/example`). Running

    go test ./...

runs all tests beginning with `fpdf_test`, which removes all PDF files from the `pdf` directory.

@jelmersnoeck The test of your package will generate a PDF in the `gofpdf/pdf` directory. It will locate the relative path automatically.

jelmersnoeck:
@jung-kurt updated it with the example package. Thanks for that, it makes it lots easier!

@jung-kurt @stanim just wanted to thank you both as well, as a beginner to the Go language, this is quite interesting and educational!

jung-kurt:
Thanks, @jelmersnoeck -- I just committed your contribution with a small aspect ratio tweak.

jelmersnoeck:
Cool!

Sorry for the clutter (the extra commits). I forgot to squash it into one commit this morning.

jung-kurt:
I'm new to git. How does one squash commits?

stanim:
http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

@jung-kurt 
This is the task of the author the pull request. You just have to merge it after it has been squashed to a single commit.

jung-kurt:
Very useful article -- thanks!

PR #42, opened 2015-08-29
Barcode contribution package.

jelmersnoeck:
This uses https://github.com/boombuler/barcode to put barcodes on the PDF page.

I've opted for the same approach as with images, where you register a barcode and then use the same identifier to put it on the page later on. This has the same benefit that we can re-use the barcode multiple times later on.

The only downside I see with this approach is that when you put a barcode on (type Code128 for example) with a code, and then another barcode (Type EAN13 for example) with the exact same code, the Code128 barcode will get overwritten. This concern is not only limited to the `barcodes` map, but will also be a concern for the `Fpdf.images` map.

@jung-kurt I have noticed a small issue with the internal package tests with this. If I run `go test contrib/barcode/barcode_test.go` all tests (of the barcode package) generate the PDFs properly (in the root/pdf folder). If I run `go test ./...` however, only the `httpimg` package seems to be creating PDFs, not the barcode package. I have checked, the examples do actually run, the files just don't get properly saved. The filepath etc. seems to be correct as well.

FYI, I will squash the commits together once this looks ok to you.

### DIFF
--- a/contrib/barcode/barcode.go
+++ b/contrib/barcode/barcode.go
@@ -1,0 +1,204 @@
+// Copyright (c) 2015 Jelmer Snoeck (Gmail: jelmer.snoeck)
+//
+// Permission to use, copy, modify, and distribute this software for any purpose
+// with or without fee is hereby granted, provided that the above copyright notice
+// and this permission notice appear in all copies.
+//
+// THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+// REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+// FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+// INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+// LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+// OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+// PERFORMANCE OF THIS SOFTWARE.
+
+// Package barcode provides helper methods for adding barcodes of different
+// types to your pdf document. It relies on the github.com/boombuler/barcode
+// package for the barcode creation.
+package barcode
+
+import (
+	"bytes"
+	"errors"
+	"github.com/boombuler/barcode"
+	"github.com/boombuler/barcode/codabar"
+	"github.com/boombuler/barcode/code128"
+	"github.com/boombuler/barcode/code39"
+	"github.com/boombuler/barcode/datamatrix"
+	"github.com/boombuler/barcode/ean"
+	"github.com/boombuler/barcode/qr"
+	"github.com/boombuler/barcode/twooffive"
+	"github.com/jung-kurt/gofpdf"
+	"image/jpeg"
+	"strconv"
+)
+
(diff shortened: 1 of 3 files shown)


jung-kurt:
Nice work! This package is an excellent use of the new contribution scheme. Everything looks exemplary.

The `go test ./...` is caused by a race condition. The `cleanup` function that is called by fpdf_test.go isn't invoked until your barcode PDFs have already been generated. As nice as it is to place all example PDFs in one place, this condition argues for either 1) not cleaning up old PDFs as part of the test sequence, or 2) putting PDFs in their own directories. Any thoughts?

In order to make map keys unique, could the barcode type be included in the key string?

marcusatbang:
Would it make sense if PDFs from `contrib/foo/foo_test.go` go into `pdf\foo`?

jung-kurt:
Good idea -- each package would then be responsible for its own documents. The various clean up cycles would not interfere with each other. We could, but wouldn't have to, drop the contrib prefix for document names.

jung-kurt:
Isolating each test package's output files is not sufficient to avoid the race condition when removing older files. It appears that the underlying operating system has not yet deleted files by the time the especially fast barcode example PDFs are generated. Adding a delay after the file deletion cycle keeps the newly generated PDFs files from being deleted. Also, this problem does not seem to be related to the indeterminate nature of when `init()` functions are called in imported packages (see [this discussion](http://stackoverflow.com/questions/24790175/when-is-the-init-function-in-go-golang-run)); I tried a rearrangement such that the file deletions positively occurred before and example functions were called.

It may be that a special external command should used to clean up the pdf directory contents.

jelmersnoeck:
A separate command sounds a bit tedious to me as you might forget it which could result you looking at the same PDF and not seeing that there's any errors. I like the idea of having a directory per package. If we combine that with having an `init()` function in each test package that calls `example.Cleanup()` (maybe with the package name, not too sure if it's possible in Go to automatically detect the caller) and that just cleans up that specific folder.

fpdf_test.go
```go
func init() {
  example.Cleanup("")
}
```

httpimg_test.go
```go
func init() {
  example.Cleanup("httpimg")
}
```

etc.

Regarding the keys, I've been thinking about it for a bit and that could indeed work, although I'm not a big fan of having to do `RegisterCode128("mycode")` and then later on use `Barcode("code-128-mycode")`. What I'd suggest in this case is returning that key in the register method which you can then use later on:

```go
pdf := …
key := barcode.RegisterCode128(pdf, "mycode")
barcode.Barcode(pdf, key, 0, 0, 100, 100, false)
```

jung-kurt:
Placing example PDFs in their own directory and explicitly calling `example.Cleanup()` to remove them seems like the cleanest approach. However, even that is not enough to keep quickly generated files, such as the barcode examples, from being removed. I suspect the underlying operating system queues operations like that. Go has a `(f *os.File) Sync()`, but this doesn't help with our requirement here, namely to make sure all deletions occur before proceeding.

Regarding the map key, yes, I think a return value from the registration call makes good sense. The caller never really has to examine the value of the key, only know that it uniquely identifies the barcode.

jelmersnoeck:
Oh right, I get it now regarding the files and the init() function. I'll have another think about this.

Moved the implementation (and examples) towards returning a unique key based on the barcode type and code.

jelmersnoeck:
@jung-kurt gave the example files some more thoughts, but the only solutions I can think of are either the one you suggested re. a separate command or have the `Filename()` function clean files up (https://github.com/jung-kurt/gofpdf/blob/master/internal/example/example.go#L86). This has the downside that old files might still be in your pdf folder though.

Are you happy with the rest of the package btw? Or would you like me to change some things?

jung-kurt:
Thanks, @jelmersnoeck, for a very useful and nicely prepared contribution.

I will continue to look into the best way to handle the file deletion race.

PR #44, opened 2015-09-10
Contribution packages specific interfaces.

jelmersnoeck:
Instead of forcing the type to be of gofpdf.Fpdf, we now use an
interface which could be whatever structure we want. This is useful for
third party libraries where they can define their own interface for the
PDF generator but still use these contribution packages.

Now external packages can implement their own PDF interface which
implements these methods and pass that along. Earlier, they would get
a type error.


jung-kurt:
Very elegant change -- nice work.

PR #45, opened 2015-09-25
Fpdf: add WriteAligned.

jelmersnoeck:
WriteAligned allows a user to write a single string and align it.
Previously, Write() would only write with Left alignment. Sometimes, you
want to align a specific text string to the center or the right.

HTMLBasic: add `<center>` tag.

This allows a user to center text within the HTML block. This way they
do not have to split up their input.

A new line will be created automatically before and after the centered
text.


jung-kurt:
Nice feature -- thanks!

PR #49, opened 2015-10-12
Compare example PDFs with reference copies

jung-kurt:
This change is based on @jelmersnoeck's initial PDF comparison code. In order to compare two PDFs, their internal resource catalogs must be similarly sorted and their internal creation timestamps must be the same. Methods were added to enable this. The catalog order needed to be addressed because gofpdf uses hash maps for internal resources and, when writing them to the PDF document, the non-deterministic iteration order resulted in inconsistent internal structure.

A helper function to compare PDFs was added. This is called by the example.Summary method if a reference file corresponding to an example PDF is found in the pdf/reference directory.

This change has not been tested on Mac or Windows yet, nor with versions of Go earlier than 1.5. Verification that it works in these different environments will be appreciated.

A desirable side effect of having the reference PDFs committed to the repository is that we can include links to them in the documentation.


jelmersnoeck:
I had a quick look over this (I'll have a more in-depth look later on). Just wondering if it's worth it extracting the functionality to get the keys from a map in a separate function?

jung-kurt:
> Just wondering if it's worth it extracting the functionality to get the keys from a map in a separate function?

Very possibly. The problem with using a function to return a slice of keys is that one function would have to be written for each value type. For example, one function for `map[string]foo` and another for `map[string]bar`. It's too bad that Go doesn't have a builtin that returns the keys of a hash in a slice. Easy to generalize with reflection, but that would be suboptimal.

marcusatbang:
Test fails on Windows, apparently in the first line:

```
< 0000000  25 50 44 46 2d 31 2e 33  0a 33 20 30 20 6f 62 6a  |%PDF-1.3.3 0 obj|
> 0000000  25 50 44 46 2d 31 2e 33  0d 0a 33 20 30 20 6f 62  |%PDF-1.3..3 0 ob|
< 0000010  0a 3c 3c 2f 54 79 70 65  20 2f 50 61 67 65 0a 2f  |.<</Type /Page./|
> 0000010  6a 0d 0a 3c 3c 2f 54 79  70 65 20 2f 50 61 67 65  |j..<</Type /Page|
< 0000020  50 61 72 65 6e 74 20 31  20 30 20 52 0a 2f 52 65  |Parent 1 0 R./Re|
> 0000020  0d 0a 2f 50 61 72 65 6e  74 20 31 20 30 20 52 0d  |../Parent 1 0 R.|
```

The differences are all newlines. I'm not sure why they'd be different, but I expect that to be a pain to enforce across the codebase. Perhaps another regex would be a good idea?

...

Did I really just say that?  :)

jung-kurt:
@marcusatbang, thanks for checking on Windows.

> The differences are all newlines. 

How naive I was to think the PDFs would have normalized line endings.

> Perhaps another regex would be a good idea? ... Did I really just say that? :)

Ha! Regrettably, they may be the most expedient solution here. It's simple to put in place and maybe we can think of a better solution later.

marcusatbang:
> How naive I was to think the PDFs would have normalized line endings.

Yeah...

Looking at a PDF generated by Illustrator CS6, it for some reason uses `0d` (`\r`, really old Mac standard) for newlines. The diff I pasted above shows the first using `0a` (`\n`, C and Unix standard) and the second using `0d0a` (`\r\n`, current Windows standard).

A quick googling doesn't suggest that either standard is more "correct" for PDF, so I suggest you pick one (probably `\n`) and enforce that throughout.

Remember that if you change the _length_ of the newline code in an existing PDF, offsets will no longer point to the start of the correct element so it'll be corrupted. This probably means that a regex won't work, because it won't just be the newlines that change, it'll be the offsets as well.

marcusatbang:
Wait... is git changing the line endings when it checks out?

The test now passes. I'm not sure I understand why.

jung-kurt:
> Remember that if you change the length of the newline code in an existing PDF, offsets will no longer point to the start of the correct element so it'll be corrupted. This probably means that a regex won't work, because it won't just be the newlines that change, it'll be the offsets as well.

Ouch. Good observation.

> The *ideal* solution would be code that understands the structure of a PDF and can determine whether they're equivalent regardless of being in a different order and using different IDs internally. But that would probably be a great deal of work.

Your earlier comment is looking more and more like the right way to approach this.

jung-kurt:
> Wait... is git changing the line endings when it checks out?

> The test now passes. I'm not sure I understand why.

Did you copy pdf/*.pdf to pdf/reference? If the reference files were created on Windows, all examples files should match.

jelmersnoeck:
I just tested this on mac (ref `212c47a`), no PDF failures, but I've got an internal package failure.

Output: https://gist.github.com/jelmersnoeck/d9a90632ff27c3ac1fe2

@marcusatbang I've got the same behaviour here. On `0e7f142` it [fails](https://gist.github.com/jelmersnoeck/2be8c108d582de14c941) for PDFs as well here.

marcusatbang:
Ah, right. I didn't touch a thing, just made sure I had a clean checkout.

marcusatbang:
> > The ideal solution would be code that understands the structure of a PDF and can determine whether they're equivalent regardless of being in a different order and using different IDs internally. But that would probably be a great deal of work.
>
> Your earlier comment is looking more and more like the right way to approach this.

That could conceivably be hooked off the FPDI port I've got going here... if you don't mind the fact that I'm not even within fifteen ballparks of finishing it.

jung-kurt:
> That could conceivably be hooked off the FPDI port I've got going here... if you don't mind the fact that I'm not even within fifteen ballparks of finishing it.

I'm fine with waiting.

jung-kurt:
@jelmersnoeck I've got the same behaviour here. On 0e7f142 it [fails](https://gist.github.com/jelmersnoeck/2be8c108d582de14c941) for PDFs as well here.

Problems with those two files are expected since I merged your aligned write fix. (They are the files that had missing space problems.) I commited replacements that should work properly now.

jelmersnoeck:
Oh ok, that makes sense.

Regarding multiple systems, would it be helpful to have something like Travis-CI set up? It currently supports Linux and OSX builds: http://docs.travis-ci.com/user/multi-os/

marcusatbang:
Depending on which of the four or five git for Windows tools you're using and what settings you picked on installation, there's a good chance git will rewrite text files for line ending. For normal files this is fine, but for a reference file you want to prevent it.

To prevent this confusion, a good first step would be to add a `.gitattributes` containing:
```
*.pdf    -text
```

jung-kurt:
> Regarding multiple systems, would it be helpful to have something like Travis-CI set up? 

It could be. The tests would be a lot more valuable if the generated PDFs could be compared byte-for-byte, but it's looking like that goal may be really hard to achieve.

jung-kurt:
> To prevent this confusion, a good first step would be to add a .gitattributes containing: `*.pdf    -text`

Done. Good advice.

marcusatbang:
Enforcing standard newlines regardless of the platform is still a good idea, even if it isn't sufficient to make them directly comparable.

marcusatbang:
Tests are all passing on Windows, and both the generated and reference files are consistently using `\n` even on Windows. I think there was more confusion than bug here. Panic over.

jung-kurt:
> Tests are all passing on Windows, and both the generated and reference files are consistently using \n even on Windows. 

I too just successfully tested the branch on Windows (XP, using go version 1.3) and OS X (10.7, using go version 1.5.1).

Was it making sure *.pdfs get checked out as binaries that resolved this problem?

marcusatbang:
I'm not certain, but I think so. It certainly contributed to my confusion, and that's what you don't want in tests.

marcusatbang:
My tests are on Window 10 and Go 1.4.1

jung-kurt:
Good -- it looks like we are getting good platform coverage for this branch.

marcusatbang:
I can probably find a few minutes at work tomorrow to run the test on Ubuntu and CentOS

jung-kurt:
Now that we will have these reference PDFs in the repository, it would be nice to have links to them in the README. However, when I click access a file residing on Github (for example, [basic.pdf](https://github.com/jung-kurt/gofpdf/raw/compare-reference-pdf/pdf/reference/basic.pdf)) my browser (Chrome) downloads it rather than renders it. Any insights?

jelmersnoeck:
It's not the native browser support, but you could link to the [GitHub viewer](https://github.com/jung-kurt/gofpdf/blob/compare-reference-pdf/pdf/reference/Fpdf_AddFont.pdf).

marcusatbang:
Test fails on Ubuntu.
```
$ go test
--- FAIL: ExampleFpdf_CreateTemplate (51.363065ms)
got:
< 0000060  3c 2f 4c 65 6e 67 74 68  20 34 30 33 3e 3e 0a 73  |</Length 403>>.s|
> 0000060  3c 2f 4c 65 6e 67 74 68  20 32 39 38 3e 3e 0a 73  |</Length 298>>.s|
< 00000b0  30 30 20 67 0a 71 20 25  21 46 28 66 6c 6f 61 74  |00 g.q %!F(float|
> 00000b0  30 30 20 67 0a 71 20 31  2e 30 30 30 30 20 30 20  |00 g.q 1.0000 0 |
< 00000c0  36 34 3d 31 29 20 30 20  30 20 25 21 46 28 66 6c  |64=1) 0 0 %!F(fl|
> 00000c0  30 20 31 2e 30 30 30 30  20 30 2e 30 30 30 30 20  |0 1.0000 0.0000 |
```
That's quite a big difference in the length. The longer version is my locally generated copy. Here's that object in the reference:
```
4 0 obj
<</Length 298>>
stream
0 J
0 j
7.09 w
BT /F0 16.00 Tf ET
0.784 0.392 0.000 RG
0.000 g
q 1.0000 0 0 1.0000 0.0000 0.0000 cm
/TPL1 Do Q
q 1.0000 0 0 1.0000 0.0000 -85.0394 cm
/TPL1 Do Q
q 1.4000 0 0 1.4000 0.0000 -506.8347 cm
/TPL1 Do Q
113.39 246.61 m 170.08 246.61 l S
BT 113.39 274.96 Td (Template example page 1) Tj ET

endstream
endobj
```
and here it is in my generated copy:
```
4 0 obj
<</Length 403>>
stream
0 J
0 j
7.09 w
BT /F0 16.00 Tf ET
0.784 0.392 0.000 RG
0.000 g
q %!F(float64=1) 0 0 %!F(float64=1) %!F(float64=0) %!F(float64=0) cm
/TPL1 Do Q
q %!F(float64=1) 0 0 %!F(float64=1) %!F(float64=0) %!F(float64=-85.04) cm
/TPL1 Do Q
q %!F(float64=1.4) 0 0 %!F(float64=1.4) %!F(float64=0) %!F(float64=-506.8) cm
/TPL1 Do Q
113.39 246.61 m 170.08 246.61 l S
BT 113.39 274.96 Td (Template example page 1) Tj ET

endstream
endobj
```
Something wrong with the floats in the printf here https://github.com/jung-kurt/gofpdf/blob/compare-reference-pdf/template.go#L82?

jung-kurt:
@marcusatbang, good catch. That business with the floats is the last problem I would have expected. Does it occur with each test run?

jung-kurt:
@jelmersnoeck 

> It's not the native browser support, but you could link to the [GitHub viewer](https://github.com/jung-kurt/gofpdf/blob/compare-reference-pdf/pdf/reference/Fpdf_AddFont.pdf).

That should work. I think some features may not be supported, but it is convenient for someone who wants to quickly view sample PDFs.

marcusatbang:
> @marcusatbang, good catch. That business with the floats is the last problem I would have expected. Does it occur with each test run?
It appears to be consistent, yes.

marcusatbang:
It is only within the `Fpdf_CreateTemplate.pdf` file.

jung-kurt:
What is very curious to me is that the floats that the fmt package reject (for example, %!F(float64=1.4)) have the same values of the reference file (1.0000, 1.0000,  0.0000,  0.0000; 1.0000, 1.0000,  0.0000,  -85.0394; etc). 

What version of go is being used?

marcusatbang:
Ah, that's Go 1.2.1, which is the version installed by Aptitude. That may be the problem.

jung-kurt:
Maybe, but it's nevertheless a concern.

marcusatbang:
Removing the precision from that format (replacing `%.4F` with `%F`) results in a valid file that contains `1.000000` instead of `1.0000`. It doesn't match the reference file, but it works.

jung-kurt:
Everything about this is interesting. We could remove the format precisions on that one line and recommit a new reference file, but I'm not sure I understand the implications yet.

jung-kurt:
template.go is the only source file that uses 'F' instead of 'f'. In the current documentation, 'F' is a synonym for 'f'. Can you try replacing `%.4F` with `%.4f`?

marcusatbang:
That worked. Replacing `%.4F` with `%.4f` on lines 82, 129 and 131 makes the test pass.

jung-kurt:
Prior to March 2014, `%F` was not permitted in format statements. Here is the [change log](https://code.google.com/p/go/source/detail?r=3c221d9d82a9aa4292369e819717f05d8cb6d9f3&name=go1.3beta2&path=/src/pkg/fmt/fmt_test.go).

jung-kurt:
> That worked. Replacing %.4F with %.4f on lines 82, 129 and 131 makes the test pass.

Good! I'll make the change on my branch.

marcusatbang:
Test passes on a fresh checkout. Approved.

marcusatbang:
(and to clarify my previous comment,

> Removing the precision from that format (replacing `%.4F` with `%F`) results in a valid file that contains `1.000000` instead of `1.0000`

I'd actually replaced them with `%f` and didn't realise the difference was significant)

jung-kurt:
> Test passes on a fresh checkout. Approved.

Thanks for work on this.

> I'd actually replaced them with %f and didn't realise the difference was significant)

That float specifier was an interesting wrinkle. One seldom considers that adding support for a feature (in this case, '%F') will have repercussions going forward.

marcusatbang:
Is there anything else you need to approve this for merging? It's pretty big, so I can understand being hesitant.

jung-kurt:
> Is there anything else you need to approve this for merging?

I was away from my desk for a few days, so I wanted to wait until I returned in the event some other gotchas materialized. Even though the merge is large, at least the comparison of example PDFs with reference documents is easily disabled if we need to back out.

I'm currently wondering how to handle links to the reference PDFs. Currently they are displayed using GitHub's basic viewer, but that does not support a number of features that the PDFs showcase such as transformations and bookmarks. The links can be to the raw versions of the reference files, in which case the PDFs get downloaded. This is more cumbersome for the user, but at least a full-featured reader can be used to examine the documents. Any thoughts?

marcusatbang:
The files themselves are in the repo, so the links in the README are just for easy browsing, right?

jung-kurt:
Correct. Links would allow someone who hasn't checked out the repository to take a look at the example documents.

marcusatbang:
Test still pass on Ubuntu with Go 1.2.1. I'll check Windows tonight.

jung-kurt:
@jelmersnoeck wrote:

> Regarding multiple systems, would it be helpful to have something like Travis-CI set up? It currently supports Linux and OSX builds: http://docs.travis-ci.com/user/multi-os/

Good idea -- I just registered the project with Travis CI and configured it to test on both Linux and OSX. Too bad Windows is not supported too.

PR #58, opened 2015-11-08
Compare: make `CompareBytes` exported.

jelmersnoeck:
All the "magic" of reference comparison happens within the
`CompareBytes` function.

Reading a file (with `ioutil.ReadFile`) gives you an array of bytes and
gofpdf supports writing to a byte buffer, where we can easily get the
bytes from.

Exporting `CompareBytes` would make it possible to use this core
functionality instead of having to convert the data into `io.Reader`.

### DIFF
--- a/compare.go
+++ b/compare.go
@@ -84,9 +84,9 @@ func checkBytes(pos int, sl1, sl2 []byte) (eq bool) {
 	return
 }
 
-// compareBytes compares the bytes referred to by sl1 with those referred to by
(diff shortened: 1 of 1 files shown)


jung-kurt:
One bit of ugliness we may want to change is the way `writeBytes()` pollutes stdout. At the cost of an extra parameter to `CompareBytes()` and friends, we could pass an io.Writer for the difference report.

Issue #66, opened 2016-03-05
Support for variable image dpi needs work


kentquirk:
Trying to right-justify an image -- even if you know its dpi -- is a bit of work. The imginfo.Extent() and related functions assume that the image is 72 dpi.

Here's what I had to do:

```go
	imginfo := pdf.RegisterImage(imgname, "")
	imgdpi := float64(288)
	widthInPixels := pdf.UnitToPointConvert(imginfo.Width())
	w := widthInPixels * pdf.PointToUnitConvert(72/imgdpi)
	pgw, _ := pdf.GetPageSize()
	lm, _, rm, _ := pdf.GetMargins()
	imgpos := pgw - lm - rm - w
	pdf.Image(imgname, imgpos, 0, -imgdpi, -imgdpi, true, "", 0, "")
```

The easiest fix I can see (since the Go image routines don't provide dpi info) would be to provide an imginfo.DpiExtent function that takes dpi as a parameter. 

A better fix would be to allow setting DPI on an image and having it taken into account. Perhaps imginfo could grow a dpi field, it could be set to 72 by default, but if you override it, it would get taken into account.

I'd be willing to submit a PR with a fix, but I'd like to know what sort of fix you'd approve before I do it, please.

jung-kurt:
@kentquirk Thanks for your offer to help with this.

> A better fix would be to allow setting DPI on an image and having it taken into account. Perhaps imginfo could grow a dpi field, it could be set to 72 by default, but if you override it, it would get taken into account.

This definitely is the right approach. A PR will be very welcome.

kentquirk:
Great, thanks. I'm on it.

jung-kurt:
Thanks for your work on this, @kentquirk 

Closing this issue because it is resolved by your [pull request](https://github.com/jung-kurt/gofpdf/pull/68).

PR #68, opened 2016-03-07
Support reading and/or setting image dpi

kentquirk:
Addresses #66. Adds the ability to support reading dpi from PNG images, and setting
dpi on any image directly; this allows images to be displayed at the
designed size.

This PR provides backwards-compatible support for reading DPI from images that support it, and for setting DPI manually and then having it properly taken into account when calculating image size.

The changes could have been much simpler except for backwards compatibility. There were two major problems:

* The parameters for the Image call. The system uses 96 dpi if the w/h parameters to Image are 0,0, but the assumption in the calculation of `ImageInfoType.Width()`, etc is 72. The only way I've come up with to preserve both is to use w/h parameters of -1 to mean "use the dpi stored in the ImageInfoType". This kind of sucks but since a resolution of 1 pixel per inch is pretty unreasonable, it will be fine. (And frankly, if anyone needs that kind of resolution to print a billboard or something, w and h are both float64, so use -1.000001 or something and it'll be fine.)
* More troublesome was the code to automatically load the DPI information from a PNG file if it exists (it's an optional chunk in the PNG specification). However, this will break applications that were relying on the fact that it wasn't being used, which will basically be everyone who loads a PNG file without specifying DPI. 

What I did was create new functions: `ImageOptions`, `RegisterImageOptions` and `RegisterImageOptionsReader` where the tp parameter was replaced with an options parameter for additional options. It currently contains two fields -- `ImageType`, which replaces tp, and `ReadDpi`, which instructs the PNG reader to read the DPI chunk from the image file.

I have noted `Image`, `RegisterImage`, and `RegisterImageReader` as deprecated -- they just call the new functions now. 

One thing this doesn't currently have is the ability to read the DPI information from JPG files. This isn't possible with the default Go JPG library (or with the PNG library, but as fpdf was already reading raw PNG files it was easy to add it). I could write a utility function to do it if that seems desirable, but it seemed a bit beyond the scope of this PR. 

We could also go further and make the options object take additional parameters that would make the Image API a little cleaner.

### DIFF
--- a/def.go
+++ b/def.go
@@ -71,6 +71,7 @@ type ImageInfoType struct {
 	dp    string
 	trns  []int
 	scale float64 // document scaling factor
+	dpi   float64
 }
 
 // PointConvert returns the value of pt, expressed in points (1/72 inch), as a
@@ -97,17 +98,26 @@ func (f *Fpdf) UnitToPointConvert(u float64) (pt float64) {
 // Extent returns the width and height of the image in the units of the Fpdf
 // object.
 func (info *ImageInfoType) Extent() (wd, ht float64) {
-	return info.w / info.scale, info.h / info.scale
+	return info.Width(), info.Height()
 }
 
 // Width returns the width of the image in the units of the Fpdf object.
 func (info *ImageInfoType) Width() float64 {
-	return info.w / info.scale
+	return info.w / (info.scale * info.dpi / 72)
 }
 
 // Height returns the height of the image in the units of the Fpdf object.
 func (info *ImageInfoType) Height() float64 {
-	return info.h / info.scale
+	return info.h / (info.scale * info.dpi / 72)
+}
+
+// Dpi sets the dots per inch for an image
+// PNG images MAY have their dpi set automatically, if the image
+// specifies it. DPI information is not currently available
+// automatically for JPG and GIF images, so if it's important to
+// you, you can set it here. It defaults to 72 dpi.
+func (info *ImageInfoType) SetDpi(dpi float64) {
+	info.dpi = dpi
 }
 
 type fontFileType struct {
--- a/fpdf.go
+++ b/fpdf.go
@@ -2250,6 +2250,16 @@ func (f *Fpdf) imageOut(info *ImageInfoType, x, y, w, h float64, flow bool, link
 		w = -96
 		h = -96
 	}
+	if w == -1 {
+		// Set image width to whatever value for dpi we read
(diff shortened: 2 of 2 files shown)


jung-kurt:
@kentquirk Excellent work and description. I will study your changes and if any questions come up I will raise them here.

@stanim You've been close to the image code in this package. Can you provide your comments?

jung-kurt:
@kentquirk 

> More troublesome was the code to automatically load the DPI information from a PNG file if it exists (it's an optional chunk in the PNG specification). However, this will break applications that were relying on the fact that it wasn't being used, which will basically be everyone who loads a PNG file without specifying DPI.

Do you have any estimate of the percentage of PNGs out there that have a DPI chunk that indicates a DPI other than 72 DPI?

kentquirk:
I only have an estimate. I searched for a few common terms on the web (like "dog") and saved the first couple of PNG files that came up for each. I have a mix of photography and designed graphics. 

There were 26 files total; 14 of them had no DPI chunk at all. Of the remaining 12, 6 were 72 dpi, 2 were 80, 2 were 300, 1 was 600, and 1 was 96. 

So the answer to your question would be 6/26, or roughly one quarter.

jung-kurt:
@kentquirk Thanks for researching that -- nice technique.

Am I right in assuming that all calls to the deprecated functions (Image(), RegisterImage(), and RegisterImageReader()) will behave exactly as before regardless of how the width and height are specified?

kentquirk:
@jung-kurt, that is the explicit intention and my tests, although somewhat limited, seem to confirm it. The whole design behind this PR was not to break backwards compatibility. I believe I have succeeded.

jung-kurt:
Excellent. Thanks @kentquirk for a well implemented enhancement. I'll merge this with a few small documentation tweaks.

kentquirk:
You're welcome. Sorry about the docs/leftover Printf; forgot to clean up at the end there. I'd have fixed it for you if you had asked. Anyway, thanks for merging!

jung-kurt:
The fact that this project mostly didn't support image DPI from the start, and bungled the little that it _did_ support, couldn't have made things any easier for you.

At some point, I would like to add an example in gofpdf_test.go to demonstrate using the new image options. In particular, it would be good place some PNGs that include DPI information (with different DPI values) and one that doesn't.

kentquirk:
I'll try to find time to do that some evening soon.

PR #73, opened 2016-05-15
Add support for Pdf417 barcodes

ruudk:
This adds support for Pdf417 barcodes using https://github.com/ruudk/golang-pdf417 :)


jung-kurt:
Thanks, @ruudk !

PR #77, opened 2016-06-30
add license title

waldyrious:
It's not strictly required, but it's useful metadata, and part of the recommended license template text (see http://choosealicense.com/licenses/isc/ and https://opensource.org/licenses/isc-license)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+ISC License
+
 Copyright (c) 2013-2016 Kurt Jung (Gmail: kurt.w.jung)
 
 Portions copyright by the contributors acknowledged in the documentation.


jung-kurt:
Good idea. I just committed the change using the MIT license. I will look into ISC license.

waldyrious:
@jung-kurt the license text [currently at the LICENSE file](https://github.com/jung-kurt/gofpdf/blob/f8baa6a71a53318060fc9dc71301cd5b06d2a803/LICENSE) *is* the ISC license. The MIT has additional clauses (which were necessary at the time it was created, but are now redundant due to the Berne convention). Compare: [MIT](http://choosealicense.com/licenses/mit/) vs [ISC](http://choosealicense.com/licenses/isc/).

jung-kurt:
Maybe, but at least the project is in good company with MIT: [Open source license usage on GitHub.com](https://github.com/blog/1964-open-source-license-usage-on-github-com)

waldyrious:
@jung-kurt definitely, feel free to use the MIT, which is a neat license on its own right :) I'm just saying the license title and the license body don't currently match, so one or the other should be changed. If you want I can make another PR adjusting that.

PR #79, opened 2016-07-28
adding support to embedded fonts as byte array

prsolucoes:
I have added support to embedded fonts (from []byte). With this you can use "https://github.com/jteeuwen/go-bindata" to create the files from fonts.

I already it with success in my project:
https://github.com/prsolucoes/godiffexporter/blob/master/main.go

Advantages:
- Everything already is on memory
- The binary can be distributed without font files

Check my Makefile to see an example of how to generate the file with embedded fonts:
https://github.com/prsolucoes/godiffexporter/blob/master/Makefile

Thanks.


jung-kurt:
Thanks, @prsolucoes, for a useful contribution. I'll merge it in when I think of the best way to demonstrate it for end-users. Probably this will be in the form of a test with sufficient  documentation to show how a Go resource file is created.

Nice work!

prsolucoes:
Hi,

So i made the tests now to you merge it.

Thanks.

jung-kurt:
Thanks @prsolucoes 

I isolated the rather large binary files to help with editing the test source file. Also, I used the calligra font to avoid copyright issues.

jung-kurt:
@prsolucoes If you would like to be acknowledged in the README, what name should I use?

prsolucoes:
Hi,

Thanks. You can use my name "Paulo Coutinho" if you want.

PR #81, opened 2016-09-08
Adding support CMYK in parsejpg

midiland:
Hello,
I'm adding a support of CMYK color in func parserjpg.


jung-kurt:
Thanks, @midiland !

Issue #84, opened 2016-10-11
Data race


RubenN:
Trying to generate QR's in pdf's in a go-routine and once in a while the following occurs:

WARNING: DATA RACE
Read at 0x00c04a9ef620 by goroutine 29:
  github.com/jung-kurt/gofpdf/contrib/barcode.Register()
      d:/projects/go/src/github.com/jung-kurt/gofpdf/contrib/barcode/barcode.go:95 +0x58
  github.com/jung-kurt/gofpdf/contrib/barcode.registerBarcode()
      d:/projects/go/src/github.com/jung-kurt/gofpdf/contrib/barcode/barcode.go:183 +0x57
  github.com/jung-kurt/gofpdf/contrib/barcode.RegisterQR()
      d:/projects/go/src/github.com/jung-kurt/gofpdf/contrib/barcode/barcode.go:161 +0xa5

Any ideas on how to prevent this?

jung-kurt:
@RubenN Is the PDF instance (returned from `New()` or `NewCustom()`) used exclusively in its own goroutine? PDF instances are definitely not shareable among goroutines.

Have you tried running your program with the [race detrector](https://blog.golang.org/race-detector)?

jung-kurt:
I should speak for myself about using the race detector. I just ran it and found three race conditions in the barcode contributed code. I will look into this and report back.

RubenN:
@jung-kurt Yes it is, I'm using the following code:

```
	key := barcode.RegisterQR(pdf, UUIDbase64url, qr.H, qr.Auto)
	barcode.Barcode(pdf, key, 258, 11, 25, 25, false)
```

As a workaround I changed it to use mutexes because of the maps beiing used in the barcode contrib:

```
	mutex.Lock()
	key := barcode.RegisterQR(pdf, UUIDbase64url, qr.H, qr.Auto)
	barcode.Barcode(pdf, key, 258, 11, 25, 25, false)
	mutex.Unlock()
```

I guess it would be best to solve this problem there though.

jung-kurt:
@RubenN Updating [boombuler/barcode](the https://github.com/boombuler/barcode) packages corrected the race problem.

It may be sufficient to run `go get -u github.com/jung-kurt/gofpdf/contrib/barcode` but, if not, you can run
 `go get -u github.com/boombuler/barcode`.

RubenN:
@jung-kurt That doesn't resolve my problem; it has something to do with access to the map in the barcode contrib:

[...]

PR #85, opened 2016-10-12
added mutex to barcode cache

RubenN:
to prevent race condition (issue #84)


jung-kurt:
Nicely done, @RubenN -- thanks!

PR #87, opened 2016-11-03
Add Times New Roman - 1251 (Rus) font

dmitry-udod:


jung-kurt:
Thanks, @dmitry-udod, for the pull request. I will decline this request because it doesn't add general value to the gofpdf package. Codepage 1251 character support is already demonstrated with the Helvetica font, and support for the examples is the only reason that fonts are included in the package at all. The instructions explain how to use external fonts, so package users are free to use any font for which a codepage descriptor can be supplied.

Issue #83, opened 2016-10-02
Example of tables with fixed height or not


flibustenet:
I begin to try gofpdf (coming from Reportlab of Python) and I need tables. Then i did some tests that I like to share to have review if I do it correctly, and if yes maybe it could go in the examples.
One table is with fixed height where the cells doesn't overlap. The other is with multi-lines with height of the highest cell.
Thanks for this lib, the speed is really impressing with thousands of pages in few seconds ! _and sorry for my english_

```Go
package main

import (
	"fmt"
	"strings"

	"github.com/jung-kurt/gofpdf"
)

const MARGECELL = 2 // marge top/bottom of cell

/*
Table with one line by row
*/
func tableClip(pdf *gofpdf.Fpdf, cols []float64, rows [][]string) {
	pagew, pageh := pdf.GetPageSize()
	_ = pagew
	mleft, mright, mtop, mbottom := pdf.GetMargins()
	_ = mleft
	_ = mright
	_ = mtop

	for _, row := range rows {
		_, lineHt := pdf.GetFontSize()
		height := lineHt + MARGECELL

		x, y := pdf.GetXY()
		// add a new page if the height of the row doesn't fit on the page
		if y+height >= pageh-mbottom {
			pdf.AddPage()
			x, y = pdf.GetXY()
		}
		for i, txt := range row {
			width := cols[i]
			pdf.Rect(x, y, width, height, "")
			pdf.ClipRect(x, y, width, height, false)
			pdf.Cell(width, height, txt)
			pdf.ClipEnd()
			x += width
		}
		pdf.Ln(-1)
	}
}

/*
Table with multi lines by row
*/
func tableMulti(pdf *gofpdf.Fpdf, cols []float64, rows [][]string) {
	pagew, pageh := pdf.GetPageSize()
	_ = pagew
	mleft, mright, mtop, mbottom := pdf.GetMargins()
	_ = mleft
	_ = mright
	_ = mtop

	for _, row := range rows {
		curx, y := pdf.GetXY()
		x := curx

		height := 0.
		_, lineHt := pdf.GetFontSize()

		for i, txt := range row {
			lines := pdf.SplitLines([]byte(txt), cols[i])
			h := float64(len(lines))*lineHt + float64(MARGECELL*len(lines))
			if h > height {
				height = h
			}
		}
		// add a new page if the height of the row doesn't fit on the page
		if pdf.GetY()+height > pageh-mbottom {
			pdf.AddPage()
[...]

PR #89, opened 2016-11-06
Add tables examples ClippedTableCells and WrappedTableCells

flibustenet:
As discussed in #83


jung-kurt:
Thanks, @flibustenet -- looks good!

PR #98, opened 2016-12-10
Added A1 and A2 page sizes

vinayaknagpal:
Added A1 and A2 page sizes


jung-kurt:
Thanks, @vinayaknagpal!

PR #101, opened 2016-12-27
Add support for Aztec barcodes

ruudk:
This adds support for Aztec barcodes from @boombuler in https://github.com/boombuler/barcode/issues/10


jung-kurt:
Thanks, @ruudk!

PR #102, opened 2017-01-04
Introduce A6-landscape

sbstjn:
I need A6 in landscape instead of portrait mode.

/cc @HenrikFricke


jung-kurt:
Hi @sbstjn. The first argument to [New()](https://godoc.org/github.com/jung-kurt/gofpdf#New) specifies whether the orientation will be portrait (default) or landscape. Unless I am missing some advantage of introducing new size specifiers for landscape mode, I am inclined to keep things as they were.

sbstjn:
Awesome, I did not notice this is already possible :)

jung-kurt:
Thanks for your interest!

PR #103, opened 2017-02-11
Accept interfaces where appropriate.

dchapes:
Also see https://github.com/mvdan/interfacer and https://www.reddit.com/r/golang/comments/5t09v5/writing_from_gonumplot_to_gofpdf_package_directly/ddm7rmm/

### DIFF
--- a/font.go
+++ b/font.go
@@ -144,22 +144,22 @@ type segmentType struct {
 	data   []byte
 }
 
-func segmentRead(f *os.File) (s segmentType, err error) {
-	if err = binary.Read(f, binary.LittleEndian, &s.marker); err != nil {
+func segmentRead(r io.Reader) (s segmentType, err error) {
+	if err = binary.Read(r, binary.LittleEndian, &s.marker); err != nil {
 		return
 	}
 	if s.marker != 128 {
 		err = fmt.Errorf("font file is not a valid binary Type1")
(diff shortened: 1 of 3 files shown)


jung-kurt:
Thanks, @dchapes -- all good points.

dchapes:
In the reddit link the poster wanted to hook together something that took an `io.Writer` and by accepting an `io.Reader` it would allow them to do so with no extra buffering with an `io.Pipe`.

An easier way for users to connect such things would be if `*Fpdf` directly implemented `io.Writer`; e.g. something as simple as:

    func (f *Fpdf) Write(p []byte) (int, error) {
	    if f.state == 2 {
		    return f.pages[f.page].Write(p)
	    }
	    return f.buffer.Write(p)
    }

However, unfortunately, there already is a `Write` method with a different signature so that would have to be renamed which, as a compatibility breaking change, would be undesirable. For future reference that's one reason why using common method names such as `Write` in a non-standard way is not a good idea and vet will complain about many of them (but for some reason not Read and Write).

jung-kurt:
Ah, what a shame I didn't appreciate that at the time. I wish `go vet` would call out non-standard uses of `Read` and `Write` to keep authors of new packages from making the same mistake.

Thanks for notes.

PR #117, opened 2017-06-10
add the ability to include JavaScript in PDFs

danmeyers:
I've been using the gofpdf library for a while now and it's so useful! 

I recently found the need to have a PDF print itself automatically on open using JavaScript. This PR is an almost verbatim re-implementation of [this FPDF extension](http://www.fpdf.org/en/script/script36.php).

I’d love to hear any feedback you might have.


jung-kurt:
Thanks, @danmeyers -- nice work!

PR #118, opened 2017-06-10
prevent panic when registering barcode fails

danmeyers:
I came across this bug when I tried to create a Code 128 barcode for non-ascii-128 string: The framework panics because the barcode it tries to register is `nil`.

I think this is a reasonable way of dealing with the bug, but @jelmersnoeck might know more, as it looks like he contributed to this file.


jung-kurt:
Good catch and fix. Thanks, @danmeyers!

PR #127, opened 2017-08-08
make barcode scaling optional, provide a method to get unscaled barcode dimensions

danmeyers:
This PR solves two problems I was having. First, I didn't know the printed size of a barcode on the PDF. Second, the barcodes printed on the PDF were scaling much too unpredictably. 

Barcode scaling works better for some kinds of barcodes than others. It would be possible using the new function to get the printed size of the barcode and use that as the scaling width. However, this solution would be subject to float arithmetic errors which could be disastrous if the unscaled width is just decimals below the minimum width necessary (see line 113 of scaledbarcode.go in boombuler/barcode). 

It seems reasonable to offer barcode scaling as an option. I suggest an implementation in this PR using pointers as optionals, a common Go practice. Happy to defer to you @jung-kurt on whether you want a separate function, however, to maintain the contract of the library as is.

### DIFF
--- a/contrib/barcode/barcode.go
+++ b/contrib/barcode/barcode.go
@@ -56,12 +56,9 @@ type barcodePdf interface {
 	SetError(err error)
 }
 
-// Barcode puts a registered barcode in the current page.
-//
-// The size should be specified in the units used to create the PDF document.
-//
-// Positioning with x, y and flow is inherited from Fpdf.Image().
-func Barcode(pdf barcodePdf, code string, x, y, w, h float64, flow bool) {
+// printBarcode internally prints the scaled or unscaled barcode to the PDF. Used by both
+// Barcode() and BarcodeUnscalable().
+func printBarcode(pdf barcodePdf, code string, x, y float64, w, h *float64, flow bool) {
 	barcodes.Lock()
 	unscaled, ok := barcodes.cache[code]
 	barcodes.Unlock()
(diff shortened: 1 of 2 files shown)


jung-kurt:
Thanks, @danmeyers, for the very useful enhancement. Everything looks really good except I am a bit reluctant to break backward compatibility. My inclination would be to replace `func Barcode` with something like `func BarcodeScale` (I'm open to a better name), and then have `Barcode` simply call `BarcodeScale` with the addresses of the width and height arguments.

Nice work -- thanks!

danmeyers:
@jung-kurt I can certainly understand the desire for backwards compatibility. I will be away for a few weeks, but happy to address this feedback upon returning. I'm hesitant to simply have `Barcode` call `BarcodeScale`, however. `BarcodeScale` takes `float64`s (as it should). This means we'd have to convert the dimensions from and then to 96 DPI which will likely result in some float arithmetic precision problems. This conversion might make the scaling factor _just_ under what it needs to be by mere decimals, which would cause an error.

We could solve this by having a shared function called by both `Barcode` and `BarcodeScale` if you're open to that, though I'm not sure what it would look like off the top of my head.

jung-kurt:
> This means we'd have to convert the dimensions from and then to 96 DPI which will likely result in some float arithmetic precision problems. 

Ah, I get that now.

> We could solve this by having a shared function called by both Barcode and BarcodeScale if you're open to that, though I'm not sure what it would look like off the top of my head.

If some code can be shared between them, that would be fine, but no problem if that's impractical.

danmeyers:
@jung-kurt Took me long enough to get to this! I made the changes we talked about in the form of a `BarcodeUnscalable` function and factored out the common barcode logic. Let me know what you think!

jung-kurt:
Thanks, @danmeyers! This is a valuable feature.

PR #131, opened 2017-08-30
Ability to write XMP Metadata

echa:
This PR adds a new function `SetXmpMetadata(xmpStream []byte)` to the API which allows the caller to insert a `/Type /Metadata /Subtype /XML` stream into the PDF file on output.

Then contents of the metadata stream must be a valid XMP metadata packet in XML format. It's up to the caller to generate such a packet, e.g. by using a library like http://github.com/echa/go-xmp. However, SetXmpMetadata does not perform any validity checks to keep the implementation lean.

I couldn't find a way to make this work with the existing API using the RawWrite* functions since they cannot be called during output (`Fpdf.enddoc()` writes PDF header, contents and trailer in one go).

### DIFF
--- a/def.go
+++ b/def.go
@@ -224,6 +224,7 @@ type Fpdf struct {
 	footerFncLpi     func(bool)                // function provided by app and called to write footer with last page flag
 	zoomMode         string                    // zoom display mode
 	layoutMode       string                    // layout display mode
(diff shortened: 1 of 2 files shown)


jung-kurt:
Thanks, @echa!

PR #132, opened 2017-08-30
Added flag to print PDF byte diffs

darrenmcc:
Currently diffs between different PDFs are printed to stdout. This is annoying, especially if the PDFs are quite large. The output is kind of useless anyway. Added a flag to enable/disable this behavior.

### DIFF
--- a/compare.go
+++ b/compare.go
@@ -75,9 +75,9 @@ func writeBytes(leadStr string, startPos int, sl []byte) {
 	fmt.Printf("|\n")
 }
 
-func checkBytes(pos int, sl1, sl2 []byte) (eq bool) {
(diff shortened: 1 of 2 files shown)


jung-kurt:
Thanks, @darrenmcc. Another approach would be to pass it an io.Writer instead of a bool. To disable the output entirely one could pass [`ioutil.Discard`](https://golang.org/pkg/io/ioutil/#pkg-variables).

If no one comments on that alternative, I'll merge this in. I'm OK with passing `false` in the call to `ComparePDFFiles()` in internal/example/example.go; that should be part of the pull request.

darrenmcc:
The nice thing about using a flag is that it can sidestep calling `writeBytes` entirely, rather than just ignoring the output.

jung-kurt:
Thanks, @darrenmcc

PR #138, opened 2017-09-29
Add "Tabloid" page size

OlegFX:
Add "Tabloid" page size, 11 x 17 inches according to ANSI X3.151.


jung-kurt:
Thanks, @OlegFX!

OlegFX:
You're welcome, @jung-kurt

PR #146, opened 2017-11-24
Replace strings.Index with strings.Contains

wgliang:
Replace with strings.Contains

Before:

if strings.Index(x, y) != -1 {}
After:

if strings.Contains(x, y) {}


jung-kurt:
Thanks, @wgliang!

PR #150, opened 2017-12-02
Add consts for page orientations

slomek:
As a new user I'd prefer to have page orientations available as consts, so that it's easier for me to get started. Also, I moved `strings.ToLower(..)` call on `orientationStr` to the top of the function so that you don't _lowerize_ a default value, and the sanitization in a single place.

### DIFF
--- a/def.go
+++ b/def.go
@@ -39,6 +39,11 @@ type gradientType struct {
 	objNum            int
 }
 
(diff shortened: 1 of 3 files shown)


jung-kurt:
Thanks @slomek -- welcome changes!

PR #158, opened 2018-01-31
Add more constants

moogle19:
Added constants for:
- Unit size
- Page size
- Border mode
- Linebreak mode
- Alignment mode


jung-kurt:
Thanks, @moogle19!

PR #160, opened 2018-02-03
sort the objs in the output more consistently

thwarted:
When CatalogSort is enabled, the sort.Strings routine is used to order
the "keys" used for the various objects being put into the PDF. Use
sort.SliceStable instead, and specify a function Less function that, if
possible, uses an attribute of the objs being sorted. This is more
deterministic.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -3199,7 +3199,7 @@ func (f *Fpdf) putfonts() {
 			fileList = append(fileList, file)
 		}
 		if f.catalogSort {
-			sort.Strings(fileList)
+			sort.SliceStable(fileList, func(i, j int) bool { return fileList[i] < fileList[j] })
 		}
 		for _, file = range fileList {
 			info = f.fontFiles[file]
@@ -3249,7 +3249,7 @@ func (f *Fpdf) putfonts() {
 			keyList = append(keyList, key)
 		}
 		if f.catalogSort {
-			sort.Strings(keyList)
+			sort.SliceStable(keyList, func(i, j int) bool { return keyList[i] < keyList[j] })
 		}
 		for _, key = range keyList {
 			font = f.fonts[key]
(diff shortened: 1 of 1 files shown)


thwarted:
This might not be the way you want to fix this, but I made a PR to start talking about it.

I have a gofpdf PDF generation code that doesn't reliably result in the same PDF object ordering in the output on multiple runs when CatalogSort is true.  This is making the tests difficult to write for this code. At first I thought my PDF generation code had some non-determinism in the order in which the PDF was drawn, but there are no branches in my code that would influence the call order of gofpdf drawing/object creation routines.

The best I can tell is that, in some cases, this is due to a combination of the use of `sort.Strings` (which uses the faster, unstable sorting function in the stdlib sort pkg) and using things like the "filename" (for fonts) or a "key name" (for other resources) to sort on.

The uses of `sort.Strings` have been replaced with calls to `sort.SliceStable` and a comparison function that compares something other than the "key" to determine ordering if possible.  For example, the width of the image is used to determine the ordering of image objects, this seemed to be more reliable. 
At first I used sort.SliceStable with a comparison function just compared the `fileList` or `keyList` entries but that produced inconsistent ordering too, but I'm not sure why.

In my PDF generation code, I don't know why the sort key being used isn't sufficiently unique to avoid dupes in the keylist that gets sorted such that a stable sort is required.

Unfortunately, I'm not yet able to share the code that produces the unreliably ordering.  I have yet to be able to produce a minimal test case showing the issue, and I'm not sure that would be super useful anyway, since the test case wouldn't fail reliably due to the use of the unstable sorting algo. I believe it shows up more often the more included fonts/images/resources are used in the PDF.

Without this change, my PDF generation code produced a different obj ordering every second or third run.
With this change, my PDF generation code produced the same output during 150 runs.

go version go1.8.4 linux/amd64

jung-kurt:
Thank you, @thwarted -- this was a good catch of a problem that has bitten me. Your use of `SliceStable()` is definitely the right approach, and your choice of catalog keys seems well reasoned and appropriate. If we ever do run into some issues with dissimilar outputs, we'll know where to look.

This is a valuable correction to the library. Nice work!

PR #176, opened 2018-04-20
EOF error when using reader

daill:
While you're using a reader for i.e. an image you'll receive an EOF from go's io package which leads to pdf creation error


jung-kurt:
Thanks, @daill!

PR #181, opened 2018-06-01
Added public wrapper "GetPageSizeStr" of private method

mrqzzz:
Add public wrapper "GetPageSizeStr" of private method
Having GetPageSizeStr public is useful for having the correct SizeType (properly scaled).
Also, that SizeType can be used to call AddPageFormat(..)


jung-kurt:
Thanks, Marcus!

mrqzzz:
:+1:

PR #186, opened 2018-07-31
barcode: Scale via the PDF rather than the barcode library

AndreRenaud:
This allows arbitrary barcode scaling, rather than whole multiple
as required by github.com/boombuler/barcode


jung-kurt:
Many thanks, @AndreRenaud!

PR #195, opened 2018-09-27
Handle URL's for images if URL is given instead of a local fs path.

9Y5:
## Context:
In php's version of fpdf, we are able to use a http URL as argument to `fpdf.Image()` like
```php
// Insert a dynamic image from a URL
$pdf->Image('http://chart.googleapis.com/chart?cht=p3&chd=t:60,40&chs=250x100&chl=Hello|World',60,30,90,0,'PNG');
```
Currently, we only can read from the local filesystem using `os.Open`.

## Summary
This commit checks if the path `fileStr` passed to `RegisterImageOptions` is a http URL, and if so, attempts to fetch the image. It otherwise defaults to the usual behaviour of reading via `os.Open`.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -34,6 +34,8 @@ import (
 	"io"
 	"io/ioutil"
 	"math"
+	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"sort"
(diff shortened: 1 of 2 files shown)


9Y5:
@jung-kurt , kindly review.

Also thanks for this library. Our team relies on it heavily.

PR #204, opened 2018-11-01
Added interface for Fpdf and Tpl

d1ngd0:
Adding an interface allows passing either a Fpdf object or a Tpl without caring which one you are getting when adding content.


jung-kurt:
Thanks, @d1ngd0.

PR #207, opened 2018-11-01
orientation of a template is now defined by the parent document.

d1ngd0:
The only issue is the CreateTemplate function has to be depricated as it doesn't have an orientation.... I created a CreateTpl function that allows a non Fpdf instantiated Tpl to be created without flipping the orientation. I made sure the broken functionality of CreateTemplate stays in place... You can decide how to handle the backwards compatibility there


jung-kurt:
Very nice work, @d1ngd0. I will augment the documentation to explain why `CreateTemplate()` is deprecated. 

The observation about the Wd - Ht placement disparity was exactly right!

PR #208, opened 2018-11-08
Added page boxes

d1ngd0:
Added a new method `SetPageBox(t string, pb PageBox)` which allows you to add an art, crop, bleed and trim box to the document. The method applies the box type to the current page as well as the default. When a new page is added, all the default boxes are applied to the page. I considered adding pageboxes as a parameter when creating a new Fpdf object, but figured it would be better not to break the api.

A PageBox struct is just a PointType and a SizeType combined.

### DIFF
--- a/def.go
+++ b/def.go
@@ -399,6 +399,7 @@ type Pdf interface {
 	SetLineWidth(width float64)
 	SetLink(link int, y float64, page int)
 	SetMargins(left, top, right float64)
+	SetPageBox(t string, pb PageBox)
 	SetProtection(actionFlag byte, userPassStr, ownerPassStr string)
 	SetRightMargin(margin float64)
 	SetSubject(subjectStr string, isUTF8 bool)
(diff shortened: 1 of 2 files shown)


jung-kurt:
Should there be some visible indication in a PDF when using page boxes? I made the following test and don't see any evidence of the page boxes in the rendered output. I have tried making page boxes both smaller and larger than the physical page but the document (as viewed with the builtin Chrome PDF viewer) doesn't look any different than if I leave the page boxes out entirely.
```go
func pagebox(pdf *gofpdf.Fpdf, boxstr string, x, y, wd, ht float64) {
	pdf.SetPageBox(boxstr, gofpdf.PageBox{gofpdf.SizeType{Wd: wd, Ht: ht}, gofpdf.PointType{X: x, Y: y}})
}

// This example demonstrates the use of a page box
func ExamplePageBox() {
	const (
		wd        = 210
		ht        = 297
		boxmargin = 5
	)
	pdf := gofpdf.New("P", "mm", "A4", "") // 210mm x 297mm
	pagebox(pdf, "bleed", 0, 0, wd, ht)
	pagebox(pdf, "trim", boxmargin, boxmargin, wd-2*boxmargin, ht-2*boxmargin)
	pdf.SetFont("Arial", "", 12)
	pdf.SetDrawColor(40, 40, 40)
	pdf.AddPage()
	pdf.Rect(0, 0, wd, ht, "D")
	pdf.Line(0, 0, wd, ht)
	pdf.Line(0, ht, wd, 0)
	fileStr := example.Filename("Fpdf_PageBox")
	err := pdf.OutputFileAndClose(fileStr)
	example.Summary(err, fileStr)
	// Output:
	// Successfully generated pdf/Fpdf_PageBox.pdf
}
```

PR #210, opened 2018-11-13
Make SplitLines and MultiCell split at the same place

### LINKED ISSUE #209: SplitLines != MultiCell
Hi,

I use SplitLines to know the max height of cells in rows of a table.

For one string I don't find the same result with SplitLines and Multicell. SplitLines say 1 line, MultiCell write 2 lines. I've looked quickly in the code of MultiCell, it seems that it doesn't use exactly the same code, something about int/float64 ?

```go
package main

import (
	"fmt"
	"os"

	"github.com/jung-kurt/gofpdf"
)

func main() {
	pdf := gofpdf.New("P", "mm", "A4", "")
	pdf.AddPage()
	pdf.SetFont("Arial", "", 8)
	lines := pdf.SplitLines([]byte("Guichon Amandine"), 26)
	fmt.Println("nb lines=", len(lines))
	_, FontSize := pdf.GetFontSize()
	y := pdf.GetY()
	pdf.MultiCell(26, FontSize, "Guichon Amandine", "", "L", false)
	yy := pdf.GetY()
	fmt.Println(FontSize, "y=", y, "yy=", yy)
	w, err := os.Create("test.pdf")
	if err != nil {
		panic(err)
	}
	defer w.Close()

	pdf.Output(w)

	fmt.Println("vim-go")
}
```

flibustenet:
This resolve #209

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2117,7 +2117,7 @@ func (f *Fpdf) SplitLines(txt []byte, w float64) [][]byte {
 	// Function contributed by Bruno Michel
 	lines := [][]byte{}
 	cw := &f.currentFont.Cw
-	wmax := (w - 2*f.cMargin) * 1000 / f.fontSize
+	wmax := int(math.Ceil((w - 2*f.cMargin) * 1000 / f.fontSize))
 	s := bytes.Replace(txt, []byte("\r"), []byte{}, -1)
 	nb := len(s)
 	for nb > 0 && s[nb-1] == '\n' {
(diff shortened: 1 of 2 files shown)


jung-kurt:
Unfortunately, the archive of the Google code repository (where the integer math was introduced on July 22, 2014) shows only a truncated commit message:

> Use integer math in SplitLines(). The math for the limit used to be done in flo

I will write Lawrence to see if he remembers the reasoning behind the change. It may be that the right thing to do is carry over the `int` and `math.Ceil()` logic to `MultiCell()` rather than undo it in `SplitLines()`.

jung-kurt:
Lawrence introduced integer math in commit e869d1301c550b65bb6b9720a5ba98423861a735. The right action will be to amend `MultiCell()` to use the same integer logic as `SplitLines()`. Do you want to look into this?

flibustenet:
yes, I will try this.

jung-kurt:
Thanks, @flibustenet -- nice work!

PR #212, opened 2018-12-04
Fixed the Pdf interface

d1ngd0:
The Pdf interface was no longer being implemented by Fpdf and Tpl due to the changes with the SetPageBox function. I have added a test to ensure Fpdf and Tpl will always implement the interface.


jung-kurt:
Good catch, @d1ngd0, and very nice test. Thanks!

PR #216, opened 2018-12-07
Added the ability to have multiple pages in a template

d1ngd0:
Templates are now aware of the number of pages they have. The main additions are the FromPage and FromPages methods which allow you to extract a Template from a template if you pass in the page number. It follows the same page mechanism as FPDF where pages start in the index 1.

Needs tests.


jung-kurt:
Thanks, @d1ngd0!

PR #219, opened 2018-12-17
Refactored Encode and Decode functions, Fixed Template ID Conflicts

d1ngd0:
- Updated Encode and Decode to ensure templates and images are pointers to the original objects
- Refactored so encoding no longer has to create p and o labels for images
- removed id from template and generate it using Bytes instead.

### DIFF
--- a/def.go
+++ b/def.go
@@ -486,8 +486,8 @@ type Fpdf struct {
 	page             int                        // current page number
 	n                int                        // current object number
 	offsets          []int                      // array of object offsets
-	templates        map[int64]Template         // templates used in this document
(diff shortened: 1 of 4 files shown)


jung-kurt:
Nice work, @d1ngd0. 

One thing we will need to be mindful of going forward is some kind of version check on independently stored templates. It's not inconceivable that breaking changes to the template structure will be made in the future. Given the short time frame templates have been saveable, it's not too big a worry now. Perhaps we should add a template format version number in the stored template so that when it is loaded back into gofpdf, the appropriate action can be taken.

PR #222, opened 2018-12-18
Updated font id generation

d1ngd0:
Generated font ID will now be the same across machines when generating the same font file.


jung-kurt:
Thanks, @d1ngd0!

PR #229, opened 2019-02-04
SVG closepath support

joewestcott:
Adds closepath (`z` or `Z` char) support to SVG handling.

closepath in the w3 SVG specification:
https://www.w3.org/TR/SVG11/paths.html#PathDataClosePathCommand


jung-kurt:
Thanks @joewestcott!

Eiskasten:
It does not. It just let the error handling not complaining about the `z` command, as it does not add it to the `[]SVGBasicSegmentType`. In addition, the calculation for the beginning coordinate is not correct as it takes always the origin coordinate and not the last `move to` instruction (from which multiple one can exist in one path) and even if the path contains only one `move to` instruction it is the wrong coordinate.

However, I fixed that in my pull request which adds support for some other instructions too. #295

PR #230, opened 2019-02-20
Add missing Go modules and add go.sum

twpayne:
`gofpdf` contains a `go.mod` file but it is missing the actual list of modules. `go.sum` is also missing. This PR fixes both problems.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,7 @@
 module github.com/jung-kurt/gofpdf
+
+require (
+	github.com/boombuler/barcode v1.0.0
+	github.com/ruudk/golang-pdf417 v0.0.0-20181029194003-1af4ab5afa58
+	golang.org/x/image v0.0.0-20190209060608-ef4a1470e0dc
+)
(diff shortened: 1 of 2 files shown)


jung-kurt:
This PR is appreciated, but we still haven't resolved the best way to handle the fact that all non-standard packages are imported only from the barcode add-on in the contrib directory. gofpdf itself without the contrib directory depends only on the standard library. There is some discussion in PR #185.

What are your recommendations?

twpayne:
Thanks for the link to #185. I'll weigh-in there.

jung-kurt:
I am now maintaining two branches: **master** (the default `go get` branch) and **v2** (the `git clone github.com/jung-kurt/gofpdf/v2` branch). v2 will comply with semantic versioning. In order to keep gofpdf free of all third party dependencies, in v2 the contrib directory (with packages for barcode and tiff support, etc) have been moved to github.com/jung-kurt/gofpdfcontrib which itself uses go.mod and go.sum and semantic versioning.

twpayne:
Belated thanks for fixing this :)

jung-kurt:
Thanks for the nudge, @twpayne -- it was definitely time for this project to get with the program and start taking semantic versioning more seriously. Incidentally, [this article](https://roberto.selbach.ca/intro-to-go-modules/) was the one I found most helpful for learning how Go modules work.

PR #231, opened 2019-02-21
Adds the ability to set word spacing.

wmatusiak:


jung-kurt:
Nice! Thanks for the feature.

Can you change the precision of the spacing from %.3f to %.5f to match the rest of the code? Also, a demonstration of this in fpdf_test.go would be really welcome. You can use the [ExampleFpdf_WriteAligned()](https://github.com/jung-kurt/gofpdf/blob/5e8bff406205400b6d067dc6d96666dedc2cf32a/fpdf_test.go#L790-L807) example as a template.

wmatusiak:
Changed as suggested.

I hope the example is OK ;)

jung-kurt:
Perfect! Thanks.

PR #232, opened 2019-02-27
Change current page of PDF document.

ykamo001:
Allows user to navigate through PDF document by changing the current page to the desired page if it's valid.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ private
 *.sublime*
 *.swp
 **/*.test
+.idea/
--- a/def.go
+++ b/def.go
@@ -442,6 +442,7 @@ type Pdf interface {
 	SetMargins(left, top, right float64)
 	SetPageBoxRec(t string, pb PageBox)
 	SetPageBox(t string, x, y, wd, ht float64)
+	SetPage(pageNum int)
 	SetProtection(actionFlag byte, userPassStr, ownerPassStr string)
 	SetRightMargin(margin float64)
 	SetSubject(subjectStr string, isUTF8 bool)
--- a/fpdf.go
+++ b/fpdf.go
@@ -383,6 +383,12 @@ func (f *Fpdf) SetPageBoxRec(t string, pb PageBox) {
 func (f *Fpdf) SetPageBox(t string, x, y, wd, ht float64) {
 	f.SetPageBoxRec(t, PageBox{SizeType{Wd: wd, Ht: ht}, PointType{X: x, Y: y}})
 }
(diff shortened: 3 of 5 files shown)


jung-kurt:
Thanks, @ykamo001. Is the idea for this to be able to revisit pages during document generation? A short example demonstrating its use would be desirable.

ykamo001:
Yes, absolutely! During document generation and editing. I'll give you the scenario and then the implementation. The reason I had to make this modification was because I was generating different graphs on different pages, but the X-axis and data points remained the same, however the Y-axis data changed. And so to save some unnecessary iterations through 2D arrays and slices, it's convenient to switch pages and make similar modifications on each page with different values.

Implementation Example:
Let's say we want to make a graph that plots Time Vs Temperature for each sensor we have.
Assume pdf has already been set up.

```
func example(pdf *gofpdf.Fpdf, time []float64,  temperaturesFromSensors[][]float64) {
        //Containers for holding the graphs and page numbers.
        var graphs []gofpdf.GridType
	var pageNums []int
        xMax := time[len(time) - 1] 
	for i , sensor := range temperaturesFromSensors {
                 //Create a new page and graph for each sensor we want to graph.
		pdf.AddPage()
		pdf.Ln(1)
                 //Custom label per sensor
		pdf.WriteAligned(0, 0, "Temperature Sensor "+strconv.Itoa(i+1)+" (\u2103) vs Time (min)", "C")
		pdf.Ln(0.5)
		graph := gofpdf.NewGrid(pdf.GetX(), pdf.GetY(), 20, 10)
		graph.TickmarksContainX(0, xMax)
                 //Custom Y axis
                 yMin := sensor[0]
                 yMax := sensor[len(sensor)-1]
		graph.TickmarksContainY(yMin, yMax)
		graph.Grid(pdf)
                //Save references and locations.
		graphs = append(graphs, graph)
		pageNums = append(pageNums, pdf.PageNo())
	}

	// For each X, graph the Y in each sensor.
	for i, currTime := range time {
		for j, sensor := range  temperaturesFromSensors{
			pdf.SetPage(pageNums[j])
			graph := graphs[j]
			temperature := sensor[i]
			pdf.Circle(graph.X(currTime), graph.Y(temperature), 0.04, "D")
		}
	}
}

```

Hope this example helps demonstrate its use! Let me know if you'd like me to add the example to the test file with some of the other examples.

jung-kurt:
Excellent! Nice work. Yes, if you could sprinkle in some example data, adding your example to fpdf_test.go would be very welcome.

jung-kurt:
Thank you very much, @ykamo001! Nice feature.

PR #244, opened 2019-04-17
Fix right alignment

Anka2310:
width has already been reduced by left and right margin, so right margin doesn't need to be subtracted twice

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2441,7 +2441,7 @@ func (f *Fpdf) WriteAligned(width, lineHeight float64, textStr, alignStr string)
 			f.Write(lineHeight, lineStr)
 			f.SetLeftMargin(lMargin)
 		case "R":
-			f.SetLeftMargin(lMargin + (width - lineWidth) - rMargin)
+			f.SetLeftMargin(lMargin + (width - lineWidth))
 			f.Write(lineHeight, lineStr)
 			f.SetLeftMargin(lMargin)
 		default:


Anka2310:
During testing I found out that automatic line breaks occur with this PR. It seems that a little more space needs to be subtracted (e.g. 5mm works), but I'm not sure what value would be best for all cases.

jung-kurt:
The [WriteAligned](https://godoc.org/github.com/jung-kurt/gofpdf#example-Fpdf-WriteAligned) example works properly without the change, but breaks after the change.

What is the situation where the current logic doesn't work properly?

Anka2310:
You can see also in the WriteAligned example that the current logic doesn't work properly; the right margin is taken into account twice, so the space to the right side is too large.

jung-kurt:
> You can see also in the WriteAligned example that the current logic doesn't work properly; the right margin is taken into account twice, so the space to the right side is too large.

Right you are, @Anka2310. It becomes quite evident when you increase the right margin in that example. Thanks for calling this out.

> It seems that a little more space needs to be subtracted (e.g. 5mm works), but I'm not sure what value would be best for all cases.

I think we need to figure out where this small discrepancy originates so we can fix it correctly.

jung-kurt:
Thanks, @Anka2310 -- this was a good catch. 

It looks like the cell margins were causing the unexpected automatic line break. This is addressed by subtracting  a "smidge" more than twice the cell margin from the string width -- not pretty, but disabling line breaks by some other means seems like it would be even less so. If you find a cleaner solution, let's use it!

PR #248, opened 2019-04-29
Add SetFontStyle method

joewestcott:
Currently, to change just the font style in user code, you have to do something like this:

````
pdf.SetFont("", "B", 0)
````

Now you can do:
````
pdf.SetFontStyle("B")
````

This implementation simply calls the exported `SetFont()` method internally. This differs slightly to `SetFontSize()`, which updates internal state and writes to the PDF in its own method. Let me know if you have any preference, or if I've missed anything.

Thanks.


jung-kurt:
Thanks, @joewestcott. This will be helpful.

PR #254, opened 2019-05-14
Add support for imported objects and templates for gofpdf v2

phpdave11:
This allows v2 of gofpdf to use gofpdi free pdf document importer.  The code that calls gofpdi is in the jung-kurt/gofpdfcontrib repository.


jung-kurt:
Many thanks, @phpdave11!

PR #259, opened 2019-05-22
maintain precision of desired w & h of barcodes

jrs526:
This fixes an issue with barcode scaling that limited
barcode sizes to integer multiples of the pdf's unit.


jung-kurt:
Thanks, @jrs526 -- useful change. And thanks for example.

PR #262, opened 2019-05-28
Added SetPage Method

SriShubham:


jung-kurt:
Thanks, @SriShubham, for the pull request but this method is already present in both active branches (master and v2). You should update your fork with `git pull` --- it's missing almost two years of enhancements and bug fixes!

PR #271, opened 2019-06-23
Improve write performance

tmcli:
Remove an unnecessary array copy. Improves performance from O(n^2) to O(n).

This assumes that each byte corresponds 1-1 to a rune.


jung-kurt:
Thanks, @tmcli! Very nice catch.

PR #279, opened 2019-07-18
fix utf8 text wrapping

seletskiy:
UTF8 text splitted correctly by SplitText function in WriteAligned
method:

https://github.com/jung-kurt/gofpdf/blob/3aaf0b1a66c0ab02d925917376898a5856061d59/fpdf.go#L2811

Then, lines of text are written via Write function, and if text is
already wrapped by given width it will be printed using 'last chunk' case:

https://github.com/jung-kurt/gofpdf/blob/3aaf0b1a66c0ab02d925917376898a5856061d59/fpdf.go#L2747-L2754

But in this case, gofpdf does not move cursor to the next cell, which
cause all text to be printed on the same line without actual wrapping.

This commit makes workaround for it by explicitly setting right margin
to force automatic line break case:

https://github.com/jung-kurt/gofpdf/blob/3aaf0b1a66c0ab02d925917376898a5856061d59/fpdf.go#L2705-L2706

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2800,8 +2800,8 @@ func (f *Fpdf) WriteLinkID(h float64, displayStr string, linkID int) {
 func (f *Fpdf) WriteAligned(width, lineHeight float64, textStr, alignStr string) {
 	lMargin, _, rMargin, _ := f.GetMargins()
 
+	pageWidth, _ := f.GetPageSize()
 	if width == 0 {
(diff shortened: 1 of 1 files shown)


jung-kurt:
Thanks, @seletskiy!

seletskiy:
@jung-kurt By mistake I've pushed wrong commit, so please revert this PR and I will make a new one.

Code in this PR is wrong.

PR #281, opened 2019-07-18
added imageoptionsreader

estenssoros:
I wanted the ability to insert images retrieved as binary (not existing locally). I wasn't sure if you wanted to be able to verify that the image is the correct extension before registering but this will be very useful to me. This is my first attempt contributing to an open-source project. Please be gentle.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2984,6 +2984,73 @@ func (f *Fpdf) ImageOptions(imageNameStr string, x, y, w, h float64, flow bool, 
 	return
 }
 
+// ImageOptionsReader registers an image, reading it from Reader r, and puts a
+// JPEG, PNG or GIF image in the current page. The size it will take on the
+// page can be specified in different ways. If both w and h are 0, the image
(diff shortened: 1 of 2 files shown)


jung-kurt:
> This is my first attempt contributing to an open-source project.

Welcome! I hope you make many more contributions. Thanks for including an example that demonstrates your work.

> I wanted the ability to insert images retrieved as binary (not existing locally). 

Image registration and handling has become seriously convoluted in gofpdf. The idea of placing an image read from an io.Reader and configured with options is a good one, but the problem is that only the options passed with the first call to `ImageOptionsReader()` would apply for subsequent renderings. (This is due to caching on the part of gofpdf. One could argue that only the image and not particular options should be cached, but that is the way the code was developed.) The fact that configured images that come from a io.Reader is something you needed indicates that better documentation is required, so I reworked your example to show how it can be done with existing methods.

Thanks for the contribution!

estenssoros:
Thanks!

Issue #272, opened 2019-06-24
Implement RoundedRect


zaddok:
A few people have noted needing `RoundedRect`, perhaps its time we include one? 

https://github.com/jung-kurt/gofpdf/issues/126

jung-kurt:
This would be suitable for a first-time contributor. Here's a PHP script on which it could be based:

[Rounded rectangle](http://www.fpdf.org/en/script/script35.php)

BeeHam:
I would like to tackle this, please.
/assign

jung-kurt:
Thanks, @BeeHam

PR #286, opened 2019-08-12
adds RoundedRect per Issue #272

BeeHam:


jung-kurt:
Thanks, @BeeHam! I straightened out one glitch in which a statement needed to be removed from a nested block. Also, I extended the example (thanks for including!) to show corner variations.

BeeHam:
Wow, thank you for the refactoring & reformatting. I don't know what happened with the formatting, I must've forgot to `go fmt`. Kind of embarrassing. I'll have to remember for next time.

jung-kurt:
Many editors have an on-write hook to which you can attach a call to `go fmt`. That way, your source never hits the disk without being in conformance.

Thanks again for contribution.

Issue #177, opened 2018-04-28
Bad data encryption


larhun:
Hi jung-kart,

running the following code generates a password protected document.  Opening the file with Adobe Acrobat (any version) does not shows the expected metadata properties (/Author, /Title,  /CreationDate, ...), except for /Producer (equal to "FPDF 1.7").

```Go
package main

import (
	"github.com/jung-kurt/gofpdf"
)

func main() {
	pdf := gofpdf.New("P", "mm", "A4", "")
	pdf.SetAuthor("author", false)
	pdf.SetTitle("title", false)
	pdf.SetProtection(0x14, "user", "owner")
	pdf.OutputFileAndClose("metadata.pdf")
}
```

This strange behavior is due to the caching of the `rc4.Cipher`, while it must be regenerated before any new encoding. Thus the function:

```Go
func (p *protectType) rc4(uint32, *[]byte)
```

should be changed so that:

- it always calls `rc4.NewCipher(key)` to get a fresh encoder,
- it caches the `key` argument not the result, for encoding data from the same indirect object.

jung-kurt:
Thanks, @larhun, for the report and required action. I'll try to get to this shortly.

PR #289, opened 2019-08-15
Add method `SetProducer`

tim-st:
This allows changing or removing the PDF producer string in the metadata of the PDF file.

I assume this will influence https://github.com/jung-kurt/gofpdf/issues/177#issue-318671952


jung-kurt:
Thanks, @tim-st!

PR #290, opened 2019-08-20
improve the speed of MultiCell

hyzgh:
Change-Id: Ieaacbf19acfce1e776eccbfa3bbc030a2ab93d5f

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2596,11 +2596,12 @@ func (f *Fpdf) MultiCell(w, h float64, txtStr, borderStr, alignStr string, fill 
 	ls := 0
 	ns := 0
 	nl := 1
+	srune := []rune(s)
 	for i < nb {
 		// Get next character
 		var c rune
 		if f.isCurrentUTF8 {
-			c = []rune(s)[i]
+			c = srune[i]
 		} else {
 			c = rune(byte(s[i]))
 		}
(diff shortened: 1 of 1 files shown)


hyzgh:
I used the MultiCell to generate a pdf, but I found it ran very slow. I had a look with the source code and found that it would convert a string slice to a rune slice during every iteration, which is the cause of the slow running speed.
To avoid that wasteful operation, I transform the string slice to a rune slice at one time and then use it in the next. With some test, I found that it actually improves the speed of MultiCell, especially for the big string.

jung-kurt:
Excellent! Many thanks, @hyzgh, for spotting that and correcting it.

joewestcott:
I've noticed repeated `[]rune(s)` [in the write method](https://github.com/jung-kurt/gofpdf/blob/648efaf66fc0b2c166ecb8b51fe68bcb0395c840/fpdf.go#L2735) also. Should be a simple fix, might even help with #301...

jung-kurt:
Great observation, @joewestcott. I wonder if it makes sense to split the string into a rune slice once at the beginning of the routine and not have to fuss with `isCurrentUTF8` at all.

joewestcott:
> I wonder if it makes sense to split the string into a rune slice once at the beginning of the routine and not have to fuss with isCurrentUTF8 at all.


I'm not certain that would work. I think the `isCurrentUTF8` distinction is important because the string given to `pdf.write()` may contain byte sequences that are not valid UTF-8 sequences. When converting to a rune slice, we're no longer indexing over bytes, but unicode code points. With bytes outside of the ASCII space (128-255) this might cause problems?

What would be nice would be to always pass gofpdf's text methods a UTF-8 string, and let it handle the codepage translation (if required) internally!

jung-kurt:
> I think the isCurrentUTF8 distinction is important because the string given to pdf.write() may contain byte sequences that are not valid UTF-8 sequences. 

You are definitely right.

> What would be nice would be to always pass gofpdf's text methods a UTF-8 string, and let it handle the codepage translation (if required) internally!

Or maybe it is time to get rid of code pages entirely in version 2 and accept only UTF-8 text.

Issue #276, opened 2019-07-03
Spaces in Chinese


achafer:
Hi, I found that when there are spaces in Chinese paragraphs, The final output style is not what I want
```go
pdf := gofpdf.New("P", "mm", "A4", "")
pdf.AddUTF8Font("notosanssc", "", "../webfonts/NotoSansSC-Regular.ttf")
pdf.AddPage()
pdf.SetFont("notosanssc", "", 16)
str := "gofpdf支持UTF-8 TrueType字体和“right-to-left”语言。" +
	"请注意，许多通用字体中可能不包含中文、日文和 韩文字符。对于这些语言，" +
	"可以使用专门的字体(例如，NotoSansSC 简体中文的字体)。"
pdf.MultiCell(0, 16, str, "", "L", false)
if err := pdf.OutputFileAndClose("hello.pdf"); err != nil {
	log.Fatal(err)
}
```
Output:

![image](https://user-images.githubusercontent.com/21477943/60585941-84943600-9dc3-11e9-88f3-3b587b2c775e.png)

jung-kurt:
Thanks for the report, @achafer. Any insights, @DarkFreedman?

jung-kurt:
@achafer, can you confirm that this works as expected now?

ArtemKor:
This space is part of glyphs. I used another font and got normal result
![anotherfont](https://user-images.githubusercontent.com/48787265/66210932-a1dae800-e6c3-11e9-951b-894a17b39ae8.png)

PR #291, opened 2019-08-21
fix split line bug

hyzgh:
In English, a word should not be split into two lines, but Chinese words can be made into two lines.
I fixed this Bug mentioned in [issue](https://github.com/jung-kurt/gofpdf/issues/276).


jung-kurt:
Thanks, @hyzgh -- nice work!

Issue #287, opened 2019-08-12
New method: pdf.SetUnderlineThickness


joewestcott:
Hiya @jung-kurt 

Do you think this method would be a useful addition to this library? It would allow the programmer to adjust the text underline thickness whilst still having it scale with the font size.
I'd be happy to create a PR with the following changes:

1. Store a new attribute on the principal struct, `userUnderlineThickness`, defaulting to `1.0`

2. Create a new method named `pdf.SetUnderlineThickness` taking a `float64`. This will set `f.userUnderlineThickness`

3. Modify `dounderline` function on line 3477 of `fpdf.go` to multiply `f.currentFont.Ut` by `f.userUnderlineThickness` (line 3 below)

````
func (f *Fpdf) dounderline(x, y float64, txt string) string {
	up := float64(f.currentFont.Up)
	ut := float64(f.currentFont.Ut) * f.userUnderlineThickness // <-- this!
	w := f.GetStringWidth(txt) + f.ws*float64(blankCount(txt))
	return sprintf("%.2f %.2f %.2f %.2f re f", x*f.k,
		(f.h-(y-up/1000*f.fontSize))*f.k, w*f.k, -ut/1000*f.fontSizePt)
}
````

Any thought, let me know. Thanks!

jung-kurt:
Go for it, @joewestcott !

PR #293, opened 2019-08-22
Add SetUnderlineThickness method

joewestcott:
As described in #287 

I've also included a test and added the result to `pdf/reference`, but it appears that folder is quite outdated compared to the current output of `go test`.

Is the reference folder meant to reflect the current output of `go test`? I can update it if so.


jung-kurt:
Nice contribution, @joewestcott -- thanks!

The reference folder started out with good intentions but new ways kept popping up that caused PDFs to be generated differently in different environments. I finally disabled the check by default -- see the call to [ComparePDFFiles](https://godoc.org/github.com/jung-kurt/gofpdf#ComparePDFFiles) in internal/example/example.go. I'm not sure whether anyone is using the reference directory in practice. It would be valuable to get a report on PDF discrepancies without causing tests to fail but I am not sure how this could be accomplished.

PR #295, opened 2019-08-28
Improve SVG support

Eiskasten:
These commits, add support for horizontal lines, vertical lines and quadratic curves. In addition this pull request implemented the closepath support truly.


Eiskasten:
I provided an example now which also shows that the `m` instruction does not work correctly. At first I thought, that this was because after each `z` instruction the new coordinates were not set but that was not the case (I committed that nevertheless, just for the formal correctness).

PR #296, opened 2019-08-30
Prevent image duplication

joewestcott:
Currently, it's possible to end up with the same image inserted into the final PDF multiple times. 

This is possible if you register the same image under two different keys (unlikely), but it's also possible when trying to use the same image in multiple templates. The [CreateTemplate example](https://github.com/jung-kurt/gofpdf/blob/5336f2c016d4b1b5e9399a84aa4657d374dd5cb3/fpdf_test.go#L2052) actually attempts this and the image is written into the PDF twice. After this  
change the filesize is reduced by about 20%.

It accomplishes this by storing the SHA-1 hash of each inserted image, and re-uses the objects ID if it's found. I've also committed my comments in a bunch of places from trying to understand the codebase.

### DIFF
--- a/def.go
+++ b/def.go
@@ -165,20 +165,20 @@ func (p PointType) XY() (float64, float64) {
 // Changes to this structure should be reflected in its GobEncode and GobDecode
 // methods.
 type ImageInfoType struct {
-	data  []byte
-	smask []byte
-	n     int
-	w     float64
-	h     float64
-	cs    string
-	pal   []byte
-	bpc   int
-	f     string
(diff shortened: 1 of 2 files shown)


joewestcott:
On second inspection, I think I've found a better way to do this. Please hold off on merging for now...!

jung-kurt:
@joewestcott  -- will do, thanks for the tip-off.

joewestcott:
I've placed my proposed breaking changes in another PR. This one should be fine!

PR #298, opened 2019-09-04
 Initialize derived templates with parent images

joewestcott:
(Based upon #296, please review that one first!)

This allows templates derived from the principal struct (`pdf.UseTemplate()` or `pdf.UseTemplateScaled()`) to inherit the parents images list. They'll be available using the same key to facilitate image re-use.

As inheritance happens with the loaded fonts and [many other settings](https://github.com/jung-kurt/gofpdf/blob/ced1e15bca902f15732b6380aaa4bdc419d87eac/template_impl.go#L278), I think images would be expected too. Due to my other PR, this shouldn't increase the final PDF size.

This one however is a ⚠️ **breaking change** ⚠️. Consider the following scenario...

1. An image is registered in the main pdf `pdf.RegisterImageReader("myKey" ...)`
2. A new template is derived using `pdf.CreateTemplate()`
3. An image being registered in the template `tpl.RegisterImageReader("myKey" ...)`

Due to the new template inheriting images, step 3 will [silently fail](https://github.com/jung-kurt/gofpdf/blob/ced1e15bca902f15732b6380aaa4bdc419d87eac/fpdf.go#L3101) due to an image already being registered with that key.

It is quite an obscure scenario, but it's best to consider it. I'm not sure how breaking changes are managed in this library (Is that what the v2 branch is for?)

Thanks :)

### DIFF
--- a/def.go
+++ b/def.go
@@ -165,20 +165,20 @@ func (p PointType) XY() (float64, float64) {
 // Changes to this structure should be reflected in its GobEncode and GobDecode
 // methods.
 type ImageInfoType struct {
-	data  []byte
-	smask []byte
-	n     int
-	w     float64
-	h     float64
-	cs    string
-	pal   []byte
-	bpc   int
-	f     string
-	dp    string
-	trns  []int
-	scale float64 // document scaling factor
-	dpi   float64
-	i     string
+	data  []byte  // Raw image data
+	smask []byte  // Soft Mask, an 8bit per-pixel transparency mask
+	n     int     // Image object number
+	w     float64 // Width
+	h     float64 // Height
+	cs    string  // Color space
+	pal   []byte  // Image color palette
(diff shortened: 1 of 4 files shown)


jung-kurt:
Thanks for the great feature, @joewestcott. I'll look it over and try to merge it in the next day or so.

> It is quite an obscure scenario, but it's best to consider it. I'm not sure how breaking changes are managed in this library

In this case, given the fairly recent addition of templates, I think the right approach is to simply document the limitation. In other cases, an additional API is usually warranted.

> (Is that what the v2 branch is for?)

The v2 branch is mostly to make gofpdf dependent only on the standard library. Templates, barcodes, and other contributed code with outside dependencies go into a separate package (gofpdfcontrib). When the dust settles on Go modules my plan is to make gofpdf/v2 the default branch.

PR #299, opened 2019-09-07
Strikeout: use "S" as styleStr

Rjinswand:
copied implementation of "underline" and set the line a little bit higher: text is striked-out

### DIFF
--- a/def.go
+++ b/def.go
@@ -542,6 +542,7 @@ type Fpdf struct {
 	fontFamily       string                     // current font family
 	fontStyle        string                     // current font style
 	underline        bool                       // underlining flag
+	strikeout        bool                       // strike out flag
 	currentFont      fontDefType                // current font info
 	fontSizePt       float64                    // current font size in points
 	fontSize         float64                    // current font size in user unit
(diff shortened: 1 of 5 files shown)


jung-kurt:
Thanks @Rjinswand! Excellent feature.

Quick question about the modified import lines ("github.com/jung-kurt/gofpdf" -> "gofpdf"). Is this simplification made possible by the `replace github.com/jung-kurt/gopdf => ./` line in go.mod?

jung-kurt:
`go vet -all` (run with `make check`) doesn't like the simplified import:

> imports gofpdf: malformed module path "gofpdf": missing dot in first path element

It may be best to stay with the fully qualified imports.

Rjinswand:
to be honest: I don't know ;) I'm not so familiar with the build process of go

I could compile my fork since I couldn't import the "internal" directory from an external source. With my changes it works. 

`fpdf_test.go:37:2: use of internal package github.com/jung-kurt/gofpdf/internal/example not allowed`

Rjinswand:
ah, there's a typo in go.mod: gopdf vs. gofpdf

jung-kurt:
> ah, there's a typo in go.mod: gopdf vs. gofpdf

Right you are! I'll look into this.

jung-kurt:
Thanks for contribution, @Rjinswand.

Import path adjustments have been an irritation when forking Go projects. The introduction of modules makes it a bit easier, see [When should I use the replace directive?](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) for more details.

PR #309, opened 2019-09-28
Add UTF16 support to RegisterAlias

ericsprauve:
Move the UTF16 conversion into RegisterAlias to allow external alias registration to support UTF16.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -3780,6 +3780,10 @@ func (f *Fpdf) SetJavascript(script string) {
 // document is closed.
 func (f *Fpdf) RegisterAlias(alias, replacement string) {
 	f.aliasMap[alias] = replacement
+	// Add a UTF16 mapping
+	aliasUtf16 := utf8toutf16(f.aliasNbPagesStr, false)
+	replacementUtf16 := utf8toutf16(replacement, false)
+	f.aliasMap[aliasUtf16] = replacementUtf16
 }
 
 func (f *Fpdf) replaceAliases() {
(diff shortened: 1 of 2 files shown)


jung-kurt:
In your line
```go
aliasUtf16 := utf8toutf16(f.aliasNbPagesStr, false)
```
did you mean
```go
aliasUtf16 := utf8toutf16(alias, false)
```
It seems to me a `utf8toutf16()` conversion also required in the `replaceAliases()` function.

kumarbros101:
test

ericsprauve:
You're correct, that was a typo on my part. I've corrected it in the latest push.

Since the only function that modifies `f.aliasMap` is `RegisterAlias`, my assumption is that if we encode and store both the UTF8 and UTF16 in the register function, then we wouldn't need to modify `replaceAliases` since `replaceAliases` pulls from `f.aliasMap` (which will have both encodings).

jung-kurt:
Thanks, @ericsprauve. I think this change warrants a test, preferably an example test. Can you modify [`ExampleFpdf_RegisterAlias()`](https://godoc.org/github.com/jung-kurt/gofpdf#example-Fpdf-RegisterAlias) in fpdf_test.go. Instead of
```go
pdf.SetFont("Arial", "", 12)
```
you will load the included UTF-8 font as follows:
```go
pdf.AddUTF8Font("dejavu", "BI", example.FontFile("DejaVuSansCondensed-BoldOblique.ttf"))
pdf.SetFont("dejavu", "BI", 12)
```
Run `make cov` to run the tests. This will generate documents in the pdf directory.

ericsprauve:
Thanks @jung-kurt, I just added the example test to the test suite and verified that the alias was correctly registered.

Screen Shot 2019-10-01 at 9 33 34 AM

jung-kurt:
Thanks for your work on this, @ericsprauve. I dug further into the alias replacement operation and found that strings that were converted with `utf8toutf16()` prior to assignment into the alias map caused the later replacement logic to fail. (UTF-16 strings in maps unexpectedly contain literal escape sequences.) Instead, the `utf8toutf16()` conversion is now deferred until the actual search-and-replace operation. Aliases with both UTF-8 and non-UTF-8 fonts are now demonstrated. These changes were commited in b1657c7960b94f9c1f90fa1efaf49e04f78e0f91 so your PR will be closed.

PR #318, opened 2019-10-04
Fix for changing font byte array

ArtemKor:
The situation when the library changed the used byte array of the font is fixed.


jung-kurt:
Thanks, @ArtemKor. Do you know why a copy of the font is required? Also, good catch on cleaning up the non-ASCII code point in `GenerateCutFont`.

OlegFX:
Previously font slices were loaded to GoFPDF object using pdf.AddUTF8FontFromBytes method couldn't be used for another GoFPDF object because the data in this slice changed after PDF generation. Now one font slice can be used any number of times.

ArtemKor:
This hotfix resolves this error. Now the byte array font is read-only and can be used many times.

jung-kurt:
Thank you, @ArtemKor!

PR #321, opened 2019-10-07
Add clean to Fpdf

Antonboom:
Hello!
Do you think this is enough or are there pitfalls?

Experimentally revealed that this only works with a calls
- https://golang.org/pkg/runtime/#GC
or
- https://golang.org/pkg/runtime/debug/#FreeOSMemory

Probably have to take out the generation of PDF in a separate service :)

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -686,7 +686,39 @@ func (f *Fpdf) Close() {
 	f.endpage()
 	// Close document
 	f.enddoc()
+
+	// Clear memory sources
+	f.clean()
+
 	return
+}
+
+func (f *Fpdf) clean() {
+	f.templates = nil
+	f.templateObjects = nil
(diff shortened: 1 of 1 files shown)


jung-kurt:
> Do you think this is enough or are there pitfalls?

Good question. If a call to [GC](https://golang.org/pkg/runtime/#GC) is needed, I wonder if it is sufficient to simply set the *Fpdf instance to nil and then call `GC()`.

jung-kurt:
This change causes a panic with TestIssue0316:

> panic: assignment to entry in nil map [recovered]
>        panic: assignment to entry in nil map

I will close this PR until we know for sure that this kind of garbage collection hint can be done safely and that the initial problem is not, in fact, the calling service's problem. Feel free to reopen if you can show otherwise. Your interest is appreciated.

PR #331, opened 2019-11-01
Add SetTextRenderingMode function

nickjwhite:
This follows section 9.3.6 of the PDF32000_2008 specification.

I use this to create "invisible" text for overlaying on images, to make searchable PDFs from OCR results.

I also made a variant branch of this with consts for the different possible modes, addtextrenderingmode-withconsts. I wasn't sure which would be a more fpdf way of doing it, so made both versions.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -2226,6 +2226,20 @@ func (f *Fpdf) SetWordSpacing(space float64) {
 	f.out(sprintf("%.5f Tw", space*f.k))
 }
 
+// SetTextRenderingMode sets the rendering mode of following text.
+// The mode can be as follows:
+// 0: Fill text
+// 1: Stroke text
+// 2: Fill, then stroke text
+// 3: Neither fill nor stroke text (invisible)
+// 4: Fill text and add to path for clipping
+// 5: Stroke text and add to path for clipping
+// 6: Fills then stroke text and add to path for clipping
+// 7: Add text to path for clipping
+func (f *Fpdf) SetTextRenderingMode(mode int) {
+	f.out(sprintf("%d Tr", mode))
+}
+
 // SetAcceptPageBreakFunc allows the application to control where page breaks
 // occur.
 //


jung-kurt:
Nice! 

Is a certain minimum document version required to use this? If so, you will need make sure the `.pdfVersion` field is properly updated; see [this code](https://github.com/jung-kurt/gofpdf/blob/9b6b7f4e784dd67fac27ae8fd4aea38136afa1c5/fpdf.go#L4601-L4603).

nickjwhite:
Good question, I hadn't thought of that!

I just did some searching, and it has been available since at least PDF version 1.3; it's documented in Section 5.2.5 of https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf

I just thought, should I add a test for this? It would have to be of the type: generate a document with different text rendering modes, compare to a pregenerated golden copy. At least I can't think of another way to test it. That said, it's simple enough that I can't really imagine anything going wrong with it that testing would catch... (famous last words). I'm happy to write up a test and add it, if you think that would be worthwhile, though.

jung-kurt:
> I just did some searching, and it has been available since at least PDF version 1.3

Good -- no need for a version check then. (1.3 is the default minimum required version for gofpdf.)

> I just thought, should I add a test for this? It would have to be of the type: generate a document with different text rendering modes, compare to a pregenerated golden copy. 

One would think that gofpdf would generate, for a given code sequence, exact copies of a document suitable for comparison. But after fussing with embedded timestamps and stable-sorted catalogs, subtle differences between Go versions and platforms _still_ sometimes results in different output. So in this case, I am fine merging this without a test.

This is a welcome feature -- thanks for the PR. I'll get to it soon.

nickjwhite:
Great, thanks @jung-kurt, I'm happy it's a welcome feature.

And thanks so much for gofpdf in general, I just recently found it and it's great :) I was deciding between this and the unipdf package, but chose gofpdf because it looked nicely designed to be easy to hack on, and I'd prefer a proper community driven project, and I'm glad I did.

jung-kurt:
Thanks, @nickjwhite. I wound up making a small example to demonstrate your routine. Feel free to extend it with the clip modes if you like.

PR #332, opened 2019-11-06
Extends support for rounded rectangles

benoitkugler:
Hi !
This PR extends the support for rounded rectangles by allowing different radius for each corner.
Two public methods are added : `RoundedRect2()` and `ClipRoundedRect2()`
It also rewrites the underlying code : the logic shared between both methods is extracted in `roundedRectPath()`, to avoid code duplication.

Any comments are welcomed; in particular, the names of the new methods could probably be more explicit. 

Thanks in advance for the review !
Benoit K.

### DIFF
--- a/fpdf.go
+++ b/fpdf.go
@@ -14,4920 +14,4929 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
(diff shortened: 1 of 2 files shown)


jung-kurt:
Thanks, @benoitkugler -- nice addition!

PR #334, opened 2019-11-07
Issue316 copy changes

jacobalberty:
This my `worsefix` branch renamed, after reflecting I believe this is the correct way to fix it as it removes any side effects from the `splice` and `assembleTables` functions in `utf8fontfile.go` in addition I added a test `TestIssue0316`, if you back out the previous fix by changing all instances of `ReadLikeCopy` to just `Read` you can see the test fails, if you return them or apply this branch then the test passes. 

I did go through and check for any other instances of append modifying the underlying arrays. There were a couple I thought might be an issue but once I sat down to inspect them I was able to confirm they were working only on new arrays.

### DIFF
--- a/fpdf_test.go
+++ b/fpdf_test.go
@@ -2813,3 +2813,27 @@ func ExampleFpdf_SetTextRenderingMode() {
 	// Output:
 	// Successfully generated pdf/Fpdf_TextRenderingMode.pdf
 }
+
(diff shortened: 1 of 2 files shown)


jung-kurt:
Many thanks, @jacobalberty -- your good thoughtful work and concise test and changes are really appreciated.

PR #335, opened 2019-11-07
Add support for embedded files

benoitkugler:
Hi, 
This PR adds support for embedded files, both at the root level via `SetAttachments()` or as page annotations via `AddAttachmentAnnotation()`. 

Summary of the changes :
 - The main content is in attachments.go 
 - A new type `Attachment` is added as an API convenience, and two internals slices are added (`attachments`  and `pageAttachments`). 
- The exported method Embed() (on Attachment) allows re-use of the same content for several annotations, without duplications.
 - File annotations require to write an object before writing the page dict. Thus putpages() is slightly modified : the assumption on object number for /Kids in putpages() would otherwise failed. 
- Two examples are added.

Best,
BK

### DIFF
--- a/attachments.go
+++ b/attachments.go
@@ -105,7 +105,8 @@ type annotationAttach struct {
 // AddAttachmentAnnotation puts a link on the current page, on the rectangle defined
 // by `x`, `y`, `w`, `h`. This link points towards the content defined in `a`, which is embedded
 // in the document.
-// No drawing is done.
+// Note than no drawing is done by this method : a method like `Cell()` or `Rect()`
+// should be called to indicate the reader there is a link here.
 // Requiring a pointer to an Attachment avoids useless copies in the resulting pdf:
 // attachment pointing to the same data will have their content only be included once,
 // and be shared amongst all links.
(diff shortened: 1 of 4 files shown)


jung-kurt:
Thanks, @benoitkugler! This is a great feature but I encounter problems in my environment (Chrome OS) when viewing the example files. The first is likely a limitation of the Chrome native viewer: I can open the file but have no access to the embedded files. Is there a way to provide an interactive link like a bookmark to them? 

The Chrome viewer fails to load the file annotations document and reports the rather unhelpful error `Failed to load PDF document.`. Any insights?

benoitkugler:
Indeed I have the same error on Chromium Web Browser.
I will look for the issue and be back when fixed :)

benoitkugler:
So... I fixed a mistake (due to a bad usage of p.pages and p.newobj()), which means annotations shouldn't cause an error anymore.
However, it seams that Chromium is unable to render file annotations... In comparison, Mozilla PDF.js, Okular or Evince (on linux) render the file properly.
So it seems it's more a reader issue, but it maybe requires more tests ?

> Is there a way to provide an interactive link like a bookmark to them?
Well, file annotations are precisely supposed to offer a "pointer on hover" linking to the file...

jung-kurt:
Nice work, @benoitkugler! Thanks for the new feature.

PR #337, opened 2019-11-18
Prevent panic on unsupported char for MultiCell

oliverpool:
Currently if an unsupported char is inserted (like `😀`) the library panics.

I think it would be better to return an error.


jung-kurt:
Thanks @oliverpool -- this is a real improvement.

